PR #7, opened 2023-09-16
Point readers directly to the Genetic Programming article

robertdfrench:
The current link is indirect: it causes readers to land on a page which *contains* a link to the Wikipedia article on Genetic Programming. This change will take leaders directly *to* the intended article.


calebwherry:
@robertdfrench Thank you for the PR! We require a CLA for contributions, feel free to sign the CLA and re-open this PR if you would like to do so: https://www.howso.com/cla/

PR #28, opened 2023-10-29
18025: Reduces overhead for building and running queries and cleans up code

howsohazard:


src/Amalgam/entity/EntityQueryCaches.cpp
```diff
@@ -345,16 +371,12 @@ void EntityQueryCaches::GetMatchingEntities(EntityQueryCondition *cond, BitArray
...
 				}
 
-				//only select cases that have all of the correct features
-				for(auto i : cond->positionLabels)
-					sbfds.IntersectEntitiesWithFeature(i, *ents_to_compute_ptr);
-
 			#ifdef MULTITHREAD_SUPPORT
 				ConvictionProcessor<KnnNonZeroDistanceQuerySBFCache, size_t, BitArrayIntegerSet> conviction_processor(buffers.convictionBuffers,
-					buffers.knnCache, distance_transform, top_k, cond->useConcurrency);
+					buffers.knnCache, distance_transform, static_cast<size_t>(cond->maxToRetrieve), cond->useConcurrency);
```
howsoRes, line 376: is cond->maxToRetrieve is limited to be no larger than total contained entities when it's defined?

howsohazard: the check I removed was redundant -- the check is already done in all of the Find* methods in a way where doing it further up adds no value

howsoRes (approved):

PR #29, opened 2023-11-03
18064: Adds previous_result opcode, enhances while opcode, renames target_* opcodes to current_*, changes reduce opcode, MAJOR

howsohazard:
Makes it easier to code reduction operations while enhancing performance and reducing memory usage.


howsoRes (dismissed):

src/Amalgam/interpreter/InterpreterOpcodesBase.cpp
```diff
@@ -592,18 +597,39 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_WHILE(EvaluableNode *en)
...
+		for(size_t i = 1; i < ocn_size; i++)
 		{
-			if(result != nullptr && result->GetType() == ENT_CONCLUDE)
-				return RemoveConcludeFromConclusion(result, evaluableNodeManager);
+			new_result = InterpretNode(ocn[i]);
+
+			if(new_result != nullptr && new_result->GetType() == ENT_CONCLUDE)
+			{
+				//if previous result is unconsumed, free if possible
+				previous_result = GetAndClearPreviousResultInConstructionStack(0);
```
howsoRes, line 613: something's going on in this method that prevents previous_result from being assigned to a variable, only allowing direct access to it

howsohazard: fixed

src/Amalgam/interpreter/Interpreter.h
```diff
@@ -147,27 +167,60 @@ class Interpreter
...
 
-		if(constructionStackIndices.size() > 0)
-			constructionStackIndices.pop_back();
+		if(constructionStackIndicesAndUniqueness.size() > 0)
+			constructionStackIndicesAndUniqueness.pop_back();
 	}
 
 	//updates the construction index at top of the stack to the new value
-	__forceinline void SetTopTargetValueIndexInConstructionStack(double new_index)
+	//assumes there is at least one construction stack entry
```
calebwherry, line 175: Is this a safe assumption?

src/Amalgam/interpreter/InterpreterOpcodesTransformations.cpp
```diff
@@ -709,6 +709,7 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_WEAVE(EvaluableNode *en)
 
 EvaluableNodeReference Interpreter::InterpretNode_ENT_REDUCE(EvaluableNode *en)
 {
+	//TODO 18064: utilize SetTopPreviousResultInConstructionStack and update documentation and tests
```
calebwherry: This still needed?

calebwherry (dismissed):

src/Amalgam/interpreter/InterpreterOpcodesTransformations.cpp
howsohazard: Good catch -- everything is done :-)

src/Amalgam/interpreter/Interpreter.h
howsohazard: All of the functions accessing the construction stack are used in conjunction, and so it's a bit redundant from a performance and use perspective to put checks in all the functions, so instead, the validity of the construction stack is checked in the opcode once instead of repeatedly for every function, and each function states the assumption in the comments.

calebwherry (dismissed):

howsoRes (approved):

PR #31, opened 2023-11-05
18066: Refactors distance code to make way for additional feature types

howsohazard:


howsoRes (approved):

src/Amalgam/GeneralizedDistance.h
```diff
@@ -270,111 +260,43 @@ class GeneralizedDistance
...
-
-				if(compute_approximate)
-				{
-					feat_params.nominalMatchDistanceTerm.SetValue(0.0, ExactApproxValuePair::APPROX);
-
-					if(effective_p_value != 0)
-						feat_params.nominalNonMatchDistanceTerm.SetValue(weight * nominal_approximate_diff, ExactApproxValuePair::APPROX);
-					else //pValue == 0
-						feat_params.nominalNonMatchDistanceTerm.SetValue(FastPow(1.0, weight), ExactApproxValuePair::APPROX);
+					constexpr double smallest_delta = 1e-100;
```
calebwherry, line 285: Isn't there a better stdlib value we can use for this? 1e-100 doesn't seem right to me here.

howsohazard: The issue here is we don't want the smallest possible value, because that will underflow the math in nearly all use cases, but it needs to be many orders of magnitude smaller than some function of the number of unique nominal values.  Imagine an extreme situation where the nominal values are GUIDs and are entirely unpredictable, and you've trained a vast number of them.  We want this smallest delta to be smaller than 1 / num nominals without underflow or overflow of the reciprocal in 64 bit floating point.  1e-100 seems to be a pretty good value.  I could add this as a comment if you think that's helpful?

calebwherry: Explanation makes sense, I'm don't think a comment is necessary.

calebwherry (approved):

PR #39, opened 2023-12-03
18578: Improves performance by tightening inner interpreter loops and hash map functions, evens out garbage collection thresholds

howsohazard:


calebwherry (approved):

src/Amalgam/evaluablenode/EvaluableNodeManagement.cpp
```diff
@@ -13,13 +13,13 @@ Concurrency::ReadWriteMutex EvaluableNodeManager::memoryModificationMutex;
...
-	firstUnusedNodeIndex = 0;
-	executionCyclesSinceLastGarbageCollection = 0;
-}
+#ifdef MULTITHREAD_SUPPORT
+const ExecutionCycleCountCompactDelta EvaluableNodeManager::minCycleCountBetweenGarbageCollectsPerThread = 150000;
+#else
+//make the next value constant if no threads
+const
+#endif
+ExecutionCycleCountCompactDelta EvaluableNodeManager::minCycleCountBetweenGarbageCollects = 150000;
```
calebwherry, line 22: I'm not sure I understand why this was split out. It looks like the "perThread" one is not used differently than non-perThread? And they are also the same value. Useful for future changes maybe?

howsohazard: There's two main reasons:
1) Most importantly: it keeps threads and thread-named things technically out of the path for single thread compilations and
2) Less importantly: in nearly all performance tests, the function that recommends garbage collection is so small that it doesn't show up in profiling.  However, there's one performance test where that consumes 8-10% of cpu when used with lots of threads.  Breaking this apart allows for one fewer load in that tight loop.

howsoRes (approved):

PR #41, opened 2023-12-16
18683: Changes difference types in distance computations to be explicit and moves deviation parameters to be more consistent.  MAJOR

howsohazard:


src/Amalgam/GeneralizedDistance.h
```diff
@@ -18,11 +18,18 @@ class GeneralizedDistance
 {
 public:
 
+	//TODO 18683: update documentation for change in deviation location for distance queries
```
calebwherry: Remove?

src/Amalgam/GeneralizedDistance.h
```diff
@@ -748,7 +733,8 @@ class GeneralizedDistance
 		if(a_type == ENIVT_NULL || b_type == ENIVT_NULL)
 			return std::numeric_limits<double>::quiet_NaN();
 
-		if(feature_type == FDT_NOMINAL)
+		if(feature_type == FDT_NOMINAL_NUMERIC
+			|| feature_type == FDT_NOMINAL_STRING || feature_type == FDT_NOMINAL_CODE)
```
calebwherry, line 735: I would add a new __forceinline IsFeatureNominal() for this like you did with the index.

src/Amalgam/SeparableBoxFilterDataStore.h
```diff
@@ -870,21 +867,25 @@ class SeparableBoxFilterDataStore
...
 			|| feature_type == GeneralizedDistance::FDT_CONTINUOUS_STRING
 			|| feature_type == GeneralizedDistance::FDT_CONTINUOUS_CODE)
 		{
 			target_values.push_back(position_value);
 			target_value_types.push_back(position_value_type);
 
-			if(feature_type == GeneralizedDistance::FDT_NOMINAL)
+			if(feature_type == GeneralizedDistance::FDT_NOMINAL_NUMERIC
+					|| feature_type == GeneralizedDistance::FDT_NOMINAL_STRING
+					|| feature_type == GeneralizedDistance::FDT_NOMINAL_CODE)
```
calebwherry, line 881: Indentation looks off here.

src/Amalgam/SeparableBoxFilterDataStore.h
```diff
@@ -965,16 +966,16 @@ class SeparableBoxFilterDataStore
...
 					feature_params, target_value_types[i], target_values[i]);
 
-				if(FastIsNaN(feature_params.knownToUnknownDifference))
-					feature_params.knownToUnknownDifference = unknown_distance_term;
-				if(FastIsNaN(feature_params.unknownToUnknownDifference))
-					feature_params.unknownToUnknownDifference = unknown_distance_term;
+				if(FastIsNaN(feature_params.knownToUnknownDistanceTerm.difference))
+					feature_params.knownToUnknownDistanceTerm.difference = unknown_distance_term;
+				if(FastIsNaN(feature_params.unknownToUnknownDistanceTerm.difference))
+					feature_params.unknownToUnknownDistanceTerm.difference = unknown_distance_term;
```
calebwherry, line 978: Just to be sure, the assignment in each of these is correct, right? "term" getting assigned to a "difference"? It is probably right, but just wanted to be extra careful.

src/Amalgam/entity/EntityQueryBuilder.h
```diff
@@ -39,6 +39,18 @@ namespace EntityQueryBuilder
 			|| type == ENT_COMPUTE_ENTITY_KL_DIVERGENCES);
 	}
 
+	//populates deviation data for feature_params from deviation_node
+	inline void PopulateFeatureDeviationData(GeneralizedDistance::FeatureParams &feature_params, EvaluableNode *deviation_node)
```
calebwherry, line 43: If you are only altering the 'deviation' member of feature_params here, any reason to not just pass that member in instead of whole thing? It'll most likely inlined so probably doesn't matter too much. Unless having the feature_param is good for the context. In which case, looks good.

src/Amalgam/entity/EntityQueryBuilder.h
```diff
@@ -70,11 +82,14 @@ namespace EntityQueryBuilder
...
-						case GetStringIdFromNodeTypeFromString(ENT_STRING): feature_type = GeneralizedDistance::FDT_CONTINUOUS_STRING;			break;	
-						case ENBISI_code:									feature_type = GeneralizedDistance::FDT_CONTINUOUS_CODE;			break;
+						case ENBISI_nominal_numeric:						feature_type = GeneralizedDistance::FDT_NOMINAL_NUMERIC;			break;
+						case ENBISI_nominal_string:							feature_type = GeneralizedDistance::FDT_NOMINAL_STRING;				break;
+						case ENBISI_nominal_code:							feature_type = GeneralizedDistance::FDT_NOMINAL_CODE;				break;
+						case ENBISI_continuous_numeric:						feature_type = GeneralizedDistance::FDT_CONTINUOUS_NUMERIC;			break;
+						case ENBISI_continuous_numeric_cyclic:				feature_type = GeneralizedDistance::FDT_CONTINUOUS_NUMERIC_CYCLIC;	break;
+						case ENBISI_continuous_string:						feature_type = GeneralizedDistance::FDT_CONTINUOUS_STRING;			break;
+						case ENBISI_continuous_code:						feature_type = GeneralizedDistance::FDT_CONTINUOUS_CODE;			break;
+
```
calebwherry: When I stretch this out in on my screen, the tabbing if off between each of these columns (including the breaks).

src/Amalgam/entity/EntityQueryBuilder.h
```diff
@@ -87,97 +102,90 @@ namespace EntityQueryBuilder
 			[&dist_params](size_t i, bool found, EvaluableNode *en) {
 				if(i < dist_params.featureParams.size())
 				{
-					dist_params.featureParams[i].unknownToUnknownDifference = std::numeric_limits<double>::quiet_NaN();
-					dist_params.featureParams[i].knownToUnknownDifference = std::numeric_limits<double>::quiet_NaN();
```
calebwherry, line 91: Is this a change in behavior? By default aren't these items 0.0 with your recent changes? Not sure that will cause issues or not.

calebwherry: Furthermore, in next section for 'get deviations', they are set to nan. But I could just not understand how they are being used.

src/Amalgam/entity/EntityQueryBuilder.h
```diff
@@ -87,97 +102,90 @@ namespace EntityQueryBuilder
...
-									dist_params.featureParams[i].typeAttributes.maxCyclicDifference = EvaluableNode::ToNumber(ocn[0]);
+									PopulateFeatureDeviationData(dist_params.featureParams[i], ocn[0]);
+
 								if(ocn_size > 1)
-									dist_params.featureParams[i].knownToUnknownDifference = EvaluableNode::ToNumber(ocn[1]);
+									dist_params.featureParams[i].knownToUnknownDistanceTerm.difference = EvaluableNode::ToNumber(ocn[1]);
+
 								if(ocn_size > 2)
-									dist_params.featureParams[i].unknownToUnknownDifference = EvaluableNode::ToNumber(ocn[2]);
+									dist_params.featureParams[i].unknownToUnknownDistanceTerm = EvaluableNode::ToNumber(ocn[2]);
```
calebwherry: Before, these were symmetric, now one sets the difference and the other sets the term. That right?

src/Amalgam/entity/EntityQueryBuilder.h
```diff
@@ -87,97 +102,90 @@ namespace EntityQueryBuilder
...
 							}
 							else //treat as singular value
 							{
-								dist_params.featureParams[i].typeAttributes.maxCyclicDifference = EvaluableNode::ToNumber(en);
+								PopulateFeatureDeviationData(dist_params.featureParams[i], en);
```
calebwherry, line 161: This is setting an entirely different value now. The docs were a bit hard to tease out if this is correct?

src/Amalgam/SeparableBoxFilterDataStore.h
howsohazard: That is correct as it is the difference which gets turned into the distance term.

src/Amalgam/entity/EntityQueryBuilder.h
howsohazard: The feature_params reference, with upcoming PRs, will have different data structures that will need to be populated, so this is appropriate for future work.

src/Amalgam/entity/EntityQueryBuilder.h
howsohazard: Yes, the value might be a single value, a list, or, in upcoming work, a more complex data structure.  Do you think something should be improved in documentation or in code at this point?

src/Amalgam/entity/EntityQueryBuilder.h
howsohazard: Hopefully the changes help.

src/Amalgam/GeneralizedDistance.h
howsohazard: Except IsFeatureNominal requires a feature index, whereas this method does not have that, only has the feature type.

src/Amalgam/SeparableBoxFilterDataStore.h
howsohazard: That's only because GitHub doesn't respect the .editorconfig in PR reviews :-/.  The indentation makes sense at least to me when the tabs are correctly indented at 4 spaces.

src/Amalgam/entity/EntityQueryBuilder.h
howsohazard: Good catch -- fixed!

src/Amalgam/entity/EntityQueryBuilder.h
howsohazard: Deviations should be 0 by default (computes regular difference without deviations), and uncertainties should be NaN by default (when NaN, the SBFDS will populate from the data).  This is the same behavior as before, just reorganized so it looks a bit different.  I reviewed the code again and I think it's correct, but let me know if you see something that isn't.  Though once we run other tests, e.g., with Howso Engine (which we will do before this gets merged), failures will catch any issues here quickly.

src/Amalgam/GeneralizedDistance.h
calebwherry: I was thinking passing the type in same as you did with index just to have one call encapsulates the 3 types as one. Not entirely needed, just thinking of matching the other method.

src/Amalgam/entity/EntityQueryBuilder.h
calebwherry: Alright, something really weird happened. This looked all kinds of messed up when I looked on Friday... now it (and the original!) look fine... no idea what that's about. So it was always fine.

src/Amalgam/entity/EntityQueryBuilder.h
calebwherry: No, I don't think so. This was just my misunderstanding of the updates. :)

calebwherry (approved):

PR #47, opened 2023-12-26
18652: Improves performance by reducing lock contention via immediate return values, fixes rare bugs

howsohazard:
Adds ability for opcodes to return immediate types directly on the stack instead of having to allocate memory, reducing lock contention, improving performance.
Fixes some rare bugs, especially around calling concat or substr with only a single parameter which needed to be interpreted, and incorrectly was not.
Improves code consistency for opcodes to reduce likelihood of errors.


src/Amalgam/evaluablenode/EvaluableNodeManagement.h
```diff
@@ -7,74 +7,142 @@
 typedef int64_t ExecutionCycleCount;
 typedef int32_t ExecutionCycleCountCompactDelta;
 
-//describes an EvaluableNode reference and whether it is uniquely referenced
+//describes an EvaluableNode value and whether it is uniquely referenced
+//this is mostly used for actual EvaluableNode *'s, and so most of the methods are built as such
+//however, if it may contain an immediate value, then that must be checked via
```
howsoRes: incomplete comment

howsoRes (approved):

PR #49, opened 2023-12-31
18781: Improves performance of accumulation opcodes, more preparation for entity write multithreading, fixes accum type bug where list would sometimes be used instead of original type

howsohazard:


src/Amalgam/entity/Entity.cpp
```diff
@@ -896,74 +889,91 @@ void Entity::SetRoot(EvaluableNode *_code, bool allocated_with_entity_enm, Evalu
...
+	}
+	else
+	{
+		//need to update node flags if new_root is cycle free, but accum_node isn't
+		if(previous_root->GetNeedCycleCheck() && (accum_code != nullptr && !accum_code->GetNeedCycleCheck()))
+			node_flags_need_update = true;
+
+		//need to update node flags if new_root is idempotent and accum_node isn't
+		if(previous_root->GetIsIdempotent() && (accum_code != nullptr && accum_code->GetIsIdempotent()))
+			node_flags_need_update = true;
```
calebwherry: Is "GetIsIdempotent()" an expensive call? If so, this always gets called, even if GetNeedCyccleCheck() already set  node_flags_need_update to true.

Also, looks a bug, missing "!" in accum_node for idempotency check (unless the comment is wrong).

src/Amalgam/entity/Entity.cpp
```diff
@@ -896,74 +889,91 @@ void Entity::SetRoot(EvaluableNode *_code, bool allocated_with_entity_enm, Evalu
...
-
-		//free current root reference
 		evaluableNodeManager.FreeNodeReference(previous_root);
 	}
 
-	size_t num_root_labels_to_update = 0;
-	if(new_root != nullptr)
-		num_root_labels_to_update = new_root->GetNumLabels();
+	//optimistically create references for the new labels, delete them if find collisions
+	string_intern_pool.CreateStringReferences(new_labels, [](auto l) { return l.first; });
```
calebwherry, line 940: Why do you create these refs up front and then del when it collides? Why not just try to inset and if it does then create the ref? Overhead of continually calling ref creation all at once vs in the loop?

howsohazard: The way the opcode is intended to be used, most of the time all the references will be new -- if there's one or more that isn't, it triggers a more expensive heavier process, where the extra delete isn't that big of a deal.  The string references are created all at once because it is one of the two most contentious locks in the system (the other being around EvaluableNode management), and performance testing has shown significant benefits to creating the references all at once under one lock.

src/Amalgam/evaluablenode/EvaluableNodeTreeManipulation.cpp
```diff
@@ -317,50 +317,32 @@ std::string EvaluableNodeTreeManipulation::MixStrings(const std::string &a, cons
...
-		if(!replacement)
+		if(CollectLabelIndexesFromTreeAndMakeLabelNormalizationPass(en, index, checked, to_replace))
 			break;
 	}
 
 	//things have been replaced, so anything might need to be updated
 	EvaluableNodeManager::UpdateFlagsForNodeTree(en, checked);
 
-	return std::make_pair(index, true);
+	return std::make_pair(index, false);
```
calebwherry, line 345: This function was returning false before (here and line 329), but now it returns true in these cases. Just want to double check that is indeed correct. I know some of the checks flipped (Cycle) so this might be the same case at the calling sights.

src/Amalgam/entity/Entity.cpp
howsohazard: GetIsIdempotent() is super cheap -- it's a constexpr load of a bit from a location constant offset from the pointer, so 2 opcodes on most ISAs (3 for RISCy ISAs).  Good catch on the missing ! -- fixed.

src/Amalgam/evaluablenode/EvaluableNodeTreeManipulation.cpp
```diff
@@ -1073,15 +1055,14 @@ bool EvaluableNodeTreeManipulation::DoesTreeContainLabels(EvaluableNode *en, Eva
 	return false;
 }
 
-bool EvaluableNodeTreeManipulation::CollectLabelIndexesFromNormalTree(EvaluableNode *tree, EvaluableNode::LabelsAssocType &index, EvaluableNode::ReferenceSetType *checked)
+bool EvaluableNodeTreeManipulation::CollectLabelIndexesFromTree(EvaluableNode *tree, EvaluableNode::LabelsAssocType &index, EvaluableNode::ReferenceSetType *checked)
 {
-	if(tree == nullptr)
-		return false;
-
```
calebwherry, line 1080: If tree can't be null (which is why you removed this, I assume), can it be made a ref instead of a pointer? Otherwise, it is never checked for null now. Which is that is pre-condition, makes sense to make it a ref to me.

src/Amalgam/evaluablenode/EvaluableNodeTreeManipulation.cpp
```diff
@@ -1097,79 +1078,44 @@ bool EvaluableNodeTreeManipulation::CollectLabelIndexesFromNormalTree(EvaluableN
 
 		//attempt to put the label in the index
 		auto [_, inserted] = index.insert(std::make_pair(label_sid, tree));
-
-		//if label already exists
-		if(!inserted)
-			return true;
+		collected_all_label_values = inserted;
```
calebwherry, line 1081: Does this do what you want? You will only get the last label's inserted bool...

src/Amalgam/evaluablenode/EvaluableNodeTreeManipulation.cpp
howsohazard: Yes, I flipped it everywhere (and changed the documentation in the definition).  Before it returned true if it needed further follow-up, but I realized it was more clear if the function verbs were more positive, and now true means it was able to retrieve what was requested without further effort.

src/Amalgam/evaluablenode/EvaluableNodeTreeManipulation.cpp
howsohazard: This is a recursive helper function, and the caller checks to make sure it's not null, and then it checks each pointer for being not null before it recurses.  Despite the +1 extra branch the CPU has to remember, it saves time overall when there some nulls; since there's already branching happening around the recurse, it saves having to go down the call before it gets rejected.  All the null checks are still happening, they're just happening before the calls.  This saves a small but measurable amount of CPU (you'll see this pattern elsewhere on recursive tree functions).

src/Amalgam/evaluablenode/EvaluableNodeTreeManipulation.cpp
howsohazard: Yep -- that's all I wanted.  Unless you know of better syntax to get an auto pair out and be able to use a variable defined in an outer scope within the pair :-).  If there is a way (it's been a little bit since I tried to make that work), I'm all ears!

src/Amalgam/evaluablenode/EvaluableNodeTreeManipulation.cpp
calebwherry: I believe this is correct, needed to verify myself.

src/Amalgam/evaluablenode/EvaluableNodeTreeManipulation.cpp
calebwherry: Ah, I see. I misunderstood the logic in the loop.

And nope, I don't think you can bind to an existing var like that.

calebwherry (approved):

PR #51, opened 2024-01-02
18825: Adds NO_REENTRANCY_LOCKS to arm8a build

calebwherry:
* Removes potential for illegal instructions by compiling with NO_REENTRANCY_LOCKS on arm8a.
* Adds platform specific sleep needed for NO_REENTRANCY_LOCKS addition.
* Removes obsolete Platform_EnsurePreciseTiming method.
* Adds missing TODOs for commented out CMake configs and code
* Adds done message to create version script to know when that step is completed.

### DIFF
--- a/src/Amalgam/PlatformSpecific.cpp
+++ b/src/Amalgam/PlatformSpecific.cpp
@@ -1,6 +1,8 @@
 //project headers:
 #include "PlatformSpecific.h"
 
+#include "Concurrency.h"
+
 //system headers:
 #include <algorithm>
 #include <array>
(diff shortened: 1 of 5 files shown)


src/Amalgam/PlatformSpecific.cpp
```diff
@@ -307,26 +309,26 @@ void Platform_GenerateSecureRandomData(void *buffer, size_t length)
...
-bool Platform_ThreadsafeLocaltime(std::time_t time_value, std::tm **localized_time)
+void Platform_Sleep(std::chrono::microseconds sleep_time_usec)
 {
-#ifdef OS_WINDOWS
-	return localtime_s(*localized_time, &time_value) == 0; //MS swaps the values and returns the wrong thing
-#else // POSIX
-	return ::localtime_r(&time_value, *localized_time) != nullptr;
+	//std::this_thread lives in the thread header. Instead of including that for all builds, use different
+	// sleep functions when compiling single threaded builds w/o re-entrant locks
+#ifndef NO_REENTRANCY_LOCKS
```
howsohazard: Would be more robust to use:
#if defined(MULTITHREAD_SUPPORT) || defined(MULTITHREAD_INTERFACE)
instead of NO_REENTRANCY_LOCKS

calebwherry: I thought about that but had convinced myself this was better. But looking back at the concurrency header and how these are used there, I agree. Fixed.

howsohazard (approved):

PR #52, opened 2024-01-03
18843: Improves efficiency of assign and accum opcodes and fixes issues with concurrent modifications to stack variables

howsohazard:


src/Amalgam/interpreter/InterpreterOpcodesBase.cpp
```diff
@@ -705,8 +706,14 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_DECLARE(EvaluableNode *en,
 			{
 				if(need_to_interpret && cn != nullptr && !cn->GetIsIdempotent())
 				{
+				#ifdef MULTITHREAD_SUPPORT
+					Concurrency::WriteLock write_lock(*callStackMutex, std::defer_lock);
+					if(callStackMutex != nullptr && GetCallStackDepth() < callStackUniqueAccessStartingDepth)
```
calebwherry: If callStackMutex can be null, you deref in line 710 but only check for null after in line 711.

src/Amalgam/interpreter/InterpreterOpcodesBase.cpp
```diff
@@ -717,6 +724,12 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_DECLARE(EvaluableNode *en,
 				}
 				else //just insert if it doesn't exist
 				{
+				#ifdef MULTITHREAD_SUPPORT
+					Concurrency::WriteLock write_lock(*callStackMutex, std::defer_lock);
+					if(callStackMutex != nullptr && GetCallStackDepth() < callStackUniqueAccessStartingDepth)
```
calebwherry: Same here with the potential null deref and null check.

calebwherry (approved):

PR #54, opened 2024-01-12
18891: Adds end-to-end surprisal as distance to all distance opcodes and queries, MINOR

howsohazard:


src/Amalgam/GeneralizedDistance.h
```diff
@@ -194,6 +194,17 @@ class GeneralizedDistance
...
+	//surprisal in nats of each of the different distributions given the appropriate uncertainty
+	//this is equal to the nats of entropy of the distribution plus the entropy of the uncertainty
+	//in the case of Laplace, the Laplace distribution is one nat, and the mean absolute deviation is half of that,
+	//therefore the value is 1.5
+	//these values can be computed via ComputeDeviationPartLaplace(0.0, 1) for each of the corresponding methods
+	//deviations other than 1 can be used, but then the result should be divided by that deviation, yielding the same value
+	static constexpr double s_surprisal_of_laplace = 1.5;
+	static constexpr double s_surprisal_of_laplace_approx = 1.500314205;
+	static constexpr double s_surprisal_of_gaussian = 1.1283791670955126;
+	static constexpr double s_surprisal_of_gaussian_approx = 1.128615528679644;
```
calebwherry, line 206: Thoughts on keeping all the statics in one place (there are a few more in this file)? I could go either way. Looks like you have them closest to where they are used right now which works but I also like to keep track of them together so that they aren't sprinkled throughout. Fine with either way.

src/Amalgam/GeneralizedDistance.h
```diff
@@ -389,6 +414,34 @@ class GeneralizedDistance
...
 	}
 
+	//exponentiats and weights the difference term contextually based on pValue
+	//note that it has extra logic to account for extreme values like infinity, negative infinity, and 0
+	__forceinline double ContextuallyExponentiateAndWeightDifferenceTerm(double dist_term, size_t index, bool high_accuracy)
+	{
+		if(dist_term == 0.0)
+			return 0.0;
+
+		double weight = featureParams[index].weight;
```
calebwherry, line 424: Do you need to verify index is valid here? Or precondition that it should always be in range?

src/Amalgam/GeneralizedDistance.h
```diff
@@ -404,102 +457,76 @@ class GeneralizedDistance
...
-				return FastPow(deviation, weight);
-		}
-		else
-		{
-			if(high_accuracy)
-				return std::pow(deviation, pValue) * weight;
-			else
-				return FastPow(deviation, pValue) * weight;
-		}
+		return deviation;
```
calebwherry, line 466: Just so I'm not missing something: this is just returning the deviation you send in? Except for that first if. Not sure that lines up with the comment? Or maybe I'm missing something...

src/Amalgam/GeneralizedDistance.h
howsohazard: For this file, since it's so mathy, I think it's probably best closest to where they're used.  But yeah, I see both arguments.

src/Amalgam/GeneralizedDistance.h
howsohazard: All of the methods here assume index is valid -- otherwise we'd have a massive performance penalty.  If there's a way or place you think I should comment that more explicitly, let me know.

src/Amalgam/GeneralizedDistance.h
howsohazard: That's exactly what it does :-).  However, once sparse deviation matrices are added (an upcoming PR that is underway), it'll do more lookups.

src/Amalgam/GeneralizedDistance.h
calebwherry: Since it is basically everywhere, I don't think there is a good place for it. So I'll resolve this.

calebwherry (dismissed):

calebwherry (approved):

PR #56, opened 2024-01-15
18982: Invalid JSON conversion now returns NaS (not-a-string) instead of an empty string

calebwherry:
Also reduces length of some extra long function names.

### DIFF
--- a/src/Amalgam/importexport/FileSupportJSON.cpp
+++ b/src/Amalgam/importexport/FileSupportJSON.cpp
@@ -370,13 +370,13 @@ bool EvaluableNodeJSONTranslation::Store(EvaluableNode *code, const std::string 
 		return false;
 	}
 
-	std::ofstream file(resource_path);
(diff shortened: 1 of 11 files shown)


src/Amalgam/importexport/FileSupportJSON.cpp
```diff
@@ -333,7 +333,7 @@ std::string EvaluableNodeJSONTranslation::EvaluableNodeToJson(EvaluableNode *cod
 	if(EvaluableNodeToJsonStringRecurse(code, json_str, sort_keys))
 		return json_str;
 	else
-		return "";
+		return "nas";
```
howsoRes: update the comment so it says 'otherwise return nas'

calebwherry: Done.

howsohazard (dismissed):

howsohazard (approved):

PR #60, opened 2024-01-21
18866: Adds header to CAML file, MAJOR

calebwherry:
* Amalgam CAML files now have a header:
  - 4 bytes - magic number:  'c' 'a' 'm' 'l'
  - uint32_t - major version
  - uint32_t - minor version
  - uint32_t - patch version
  - Any previously created CAML files are now unsupported
* Updates LoadEntity to have richer status reporting when files cannot be read and adds VerifyEntity that verifies entity from a file but doesn't load it
* Removes CSTL (compressed strings) file support
* Adds `amlg_code` directory to files list
* Cleans-up C++ header includes
* Cleans-up test output files and moves REPL to examples folder


src/Amalgam/Amalgam.h
```diff
@@ -63,6 +62,9 @@ extern "C"
 	AMALGAM_EXPORT wchar_t *ExecuteEntityJsonPtrWide(char *handle, char *label, char *json);
 	AMALGAM_EXPORT char *ExecuteEntityJsonPtr(char *handle, char *label, char *json);
 
+	//TODO:18866 - This should return JSON with a bool and a helpful error message on error
```
howsohazard: I'd recommend moving this up to where load + store entity are, and to better future-proof it, maybe rename it to ValidateEntity(char *path).  For now, if it's not a caml, just return true if the file exists.  If it is a caml, then do the processing wrt version.

src/Amalgam/importexport/FileSupportCAML.h
```diff
@@ -0,0 +1,20 @@
+#pragma once
+
+//system headers:
+#include <iostream>
+#include <string>
+
+//TODO:18866 - should this just be a namespace a free functions? Probably...
```
howsohazard: Yep, go with namespace to match the pattern of the other file types.

src/Amalgam/importexport/FileSupportCAML.cpp
```diff
@@ -0,0 +1,124 @@
...
+		//validate version
+		std::string dev_build = AMALGAM_VERSION_SUFFIX;
+		if(!dev_build.empty())
+			return true; //dev build
+		else if(
+			(major > s_current_major) ||
+			(major == s_current_major && minor > s_current_minor) ||
+			(major == s_current_major && minor == s_current_minor && patch > s_current_patch))
+		{
+			return false; //newer version, not supported
```
howsohazard: Recommend also writing out to stderr here like the other file formats.

src/Amalgam/importexport/FileSupportCAML.h
calebwherry: Fixed.

src/Amalgam/Amalgam.h
calebwherry: Fixed.

src/Amalgam/importexport/FileSupportCAML.cpp
calebwherry: Fixed.

examples/repl/repl.amlg
```diff
@@ -0,0 +1,19 @@
+(while (true)
+	(system "printline" "> ")
+	;(print (call (parse (system "readline"))) "\n")
```
howsohazard: remove this

examples/repl/repl.amlg
```diff
@@ -0,0 +1,19 @@
...
+			(= "(" (first (explode input)) )
+			(print (call (parse input)) "\n")
+
+			;else
+			(print
+				(call
+					(parse (concat "(" input ")"))
+				)
+				"\n"
+			)
```
howsohazard: probably get rid of the if / else with the parenthesis -- that is super old and the language doesn't attempt to transform invalid opcodes anymore

src/Amalgam/AssetManager.h
```diff
@@ -108,33 +109,50 @@ class AssetManager
 
 	//loads filename into the buffer specified by b (of type BufferType of elements BufferElementType), returns true if successful, false if not
 	template<typename BufferType>
-	static bool LoadFileToBuffer(const std::string &filename, BufferType &b)
+	static bool LoadFileToBuffer(const std::string &filename, std::string& file_type, BufferType &b)
```
howsohazard: string &

src/Amalgam/entity/EntityExternalInterface.cpp
```diff
@@ -15,12 +25,15 @@ bool EntityExternalInterface::LoadEntity(std::string &handle, std::string &path,
...
+	LoadEntityStatus status = LOADENTITYSTATUS_DEFAULT;
 	std::string file_type = "";
-	Entity *entity = asset_manager.LoadEntityFromResourcePath(path, file_type, persistent, load_contained_entities, false, true, rand_seed);
+	Entity *entity = asset_manager.LoadEntityFromResourcePath(path, file_type, persistent, load_contained_entities, false, true, rand_seed, status);
 	asset_manager.SetRootPermission(entity, true);
 
-	if(entity == nullptr)
-		return false;
+	if(entity == nullptr || !status.loaded)
+	{
```
howsohazard: don't need {}'s

src/Amalgam/importexport/FileSupportCAML.cpp
```diff
@@ -0,0 +1,102 @@
...
+		return false;
+
+	return true;
+}
+
+std::tuple<std::string, std::string, bool> FileSupportCAML::ReadHeader(std::ifstream &stream, size_t &header_size)
+{
+	uint8_t magic[4] = { 0 };
+	if(!stream.read(reinterpret_cast<char *>(magic), sizeof(magic)))
+		return std::make_tuple("Cannot read magic number", "", false);
```
howsohazard: Maybe be a little more descriptive than "magic number"

howsohazard: Same below

src/Amalgam/importexport/FileSupportCAML.h
```diff
@@ -0,0 +1,16 @@
...
+//system headers:
+#include <fstream>
+#include <ostream>
+#include <string>
+#include <utility>
+
+namespace FileSupportCAML
+{
+	//read the header from the stream
+	std::tuple<std::string,std::string, bool> ReadHeader(std::ifstream &stream, size_t &header_size);
```
howsohazard: Space after ,

src/Amalgam/Amalgam.h
```diff
@@ -15,8 +16,17 @@
...
 extern "C"
 {
+	//status from LoadEntity
+	typedef struct
+	{
+		bool loaded;
+		const char *message;
+		const char *version;
+	} LoadEntityStatus;
+	#define LOADENTITYSTATUS_DEFAULT { true, "", "" }
```
howsohazard: Seems like this should be a class and use RAII instead of a macro.

src/Amalgam/AssetManager.cpp
```diff
@@ -219,14 +188,27 @@ Entity *AssetManager::LoadEntityFromResourcePath(std::string &resource_path, std
...
 		{
 			EvaluableNode **seed = metadata->GetMappedChildNode(ENBISI_rand_seed);
 			if(seed != nullptr)
 				default_random_seed = EvaluableNode::ToStringPreservingOpcodeType(*seed);
+
+			// TODO: string leak here
+			EvaluableNode **version = metadata->GetMappedChildNode(ENBISI_version);
+			if(version != nullptr)
+			{
+				auto version_str = EvaluableNode::ToStringPreservingOpcodeType(*version);
```
howsohazard: Just ToString might be more appropriate here -- the reason this version is used for the random seed is because it's trying to coerce it into a random number.

src/Amalgam/AssetManager.cpp
```diff
@@ -219,14 +188,27 @@ Entity *AssetManager::LoadEntityFromResourcePath(std::string &resource_path, std
...
+	EvaluableNode *metadata = LoadResourcePath(metadata_filename, metadata_base_path, metadata_extension, &new_entity->evaluableNodeManager, escape_filename, status);
 	if(metadata != nullptr)
 	{
 		if(EvaluableNode::IsAssociativeArray(metadata))
 		{
 			EvaluableNode **seed = metadata->GetMappedChildNode(ENBISI_rand_seed);
 			if(seed != nullptr)
 				default_random_seed = EvaluableNode::ToStringPreservingOpcodeType(*seed);
+
+			// TODO: string leak here
```
howsohazard: does the string leak still exist after garbage collection is called?  GC should clean it up.  it'd probably be ever so slightly more efficient for bulk loads if we added a new_entity->evaluableNodeManager.FreeNodeTree here and passed in the metadata, and that should get rid of the extra string reference immediately instead of waiting for GC

howsohazard (changes requested):

src/Amalgam/Amalgam.h
calebwherry: So at the C API level it can't be. It has to be pure C to go across the library boundary safely. But it is clunky to have this used in the C++ interface so I'll create a C++ one that is much easier to work with and handles resources correctly.

FWIW: this struct worked well in my first implementation but not so much after I refactored.

src/Amalgam/AssetManager.cpp
calebwherry: I'll check out the behavior with GC and see if the leak goes away.

src/Amalgam/Amalgam.h
calebwherry: Created a C++ status for cleaner usage internally with a converter at the C/C++ boundary.

src/Amalgam/AssetManager.cpp
calebwherry: Memory leak was actually legit and caused by incorrect usage of the C status struct. Moving to C++ struct fixed it.

src/Amalgam/Amalgam.h
```diff
@@ -15,8 +16,16 @@
 
 extern "C"
 {
+	//status from LoadEntity
+	typedef struct
+	{
+		bool loaded;
+		char *message;
+		char *version;
+	} LoadEntityStatus_s;
```
howsohazard: This is kind of archaic C, and given that we require C++17, I think we can require newer C and just do "struct LoadEntityStatus"...  Also, probably don't need the _s to be modern as well.

src/Amalgam/AssetManager.cpp
```diff
@@ -202,13 +171,14 @@ bool AssetManager::StoreResourcePath(EvaluableNode *code, std::string &resource_
...
+	bool persistent, bool load_contained_entities, bool escape_filename, bool escape_contained_filenames,
+	std::string default_random_seed, LoadEntityStatus &status)
 {
 	std::string resource_base_path;
 	Entity *new_entity = new Entity();
 
-	EvaluableNodeReference code = LoadResourcePath(resource_path, resource_base_path, file_type, &new_entity->evaluableNodeManager, escape_filename);
-	if(code == nullptr)
+	EvaluableNodeReference code = LoadResourcePath(resource_path, resource_base_path, file_type, &new_entity->evaluableNodeManager, escape_filename, status);
+	if(code == nullptr || !status.loaded)
```
howsohazard: probably can get rid of the "code == nullptr" such that a single null is valid -- the status.loaded will capture other errors, right?

src/Amalgam/AssetManager.h
```diff
@@ -108,33 +112,50 @@ class AssetManager
...
+		{
+			auto [error_string, version, success] = FileSupportCAML::ReadHeader(f, header_size);
+			if(!success)
+				return false;
+		}
+
 		f.seekg(0, std::ios::end);
-		b.reserve(f.tellg());
-		f.seekg(0, std::ios::beg);
+		b.reserve((std::streamoff)f.tellg() - header_size);
```
howsohazard: Should the streamoff be a static cast?

src/Amalgam/entity/EntityExternalInterface.h
```diff
@@ -11,6 +12,18 @@
 #include <string>
 #include <vector>
 
+//status from LoadEntity
+class LoadEntityStatus
+{
+public:
+	LoadEntityStatus();
+	void Set(bool loaded_in, std::string message_in = std::string(), std::string version_in = std::string());
```
howsohazard: Not sure Set is the best name?  SetStatus?  SetStatusAndVersion?

src/Amalgam/importexport/FileSupportCAML.h
```diff
@@ -0,0 +1,16 @@
...
+
+//system headers:
+#include <fstream>
+#include <ostream>
+#include <string>
+#include <utility>
+
+namespace FileSupportCAML
+{
+	//read the header from the stream
```
howsohazard, line 11: should list what the three things returned are

src/Amalgam/string/StringManipulation.cpp
```diff
@@ -69,6 +71,18 @@ std::string StringManipulation::RemoveFirstWord(std::string &str)
...
 }
 
+std::vector<std::string> StringManipulation::Split(std::string &s, char delim)
+{
+	std::vector<std::string> ret;
+	std::stringstream ss { s };
+	std::string item;
+
+	while(std::getline(ss, item, delim))
+		ret.push_back (item);
```
howsohazard: no space after push_back

src/Amalgam/Amalgam.h
calebwherry: Moving to newer C style is fine, will update. The "_s" is because I also have the C++ one that doesn't have the postfix. So best way was to use the older style naming with postfix in C one. We don't have any namespaces for the C++ ones so there will be a collision unless you have another suggestion?

src/Amalgam/AssetManager.h
calebwherry: Fixed.

src/Amalgam/entity/EntityExternalInterface.h
calebwherry: SetStatus works.

src/Amalgam/string/StringManipulation.cpp
calebwherry: Fixed.

src/Amalgam/AssetManager.cpp
calebwherry: Fixed.

src/Amalgam/AssetManager.h
```diff
@@ -110,29 +110,31 @@ class AssetManager
 		return rootEntities.find(entity) != end(rootEntities);
 	}
 
-	//loads filename into the buffer specified by b (of type BufferType of elements BufferElementType), returns true if successful, false if not
+	//loads filename into the buffer specified by b (of type BufferType of elements BufferElementType)
+	//if successful, returns empty strings and true
```
howsohazard: Wouldn't it be useful to return the version regardless even if successful?  E.g., if it's off by a point version

calebwherry: Ah, yes, it would. I'll update the places where this is called.

calebwherry: Fixed.

src/Amalgam/Amalgam.h
calebwherry: Update C struct to not have postfix and moved C++ class into another class so no name collisions.

howsohazard (dismissed):

cademack (approved):

PR #67, opened 2024-01-29
19133: Reduces contention for concurrency, fixes rare slow memory leak

howsohazard:


src/Amalgam/KnnCache.h
```diff
@@ -67,13 +68,11 @@ class KnnNonZeroDistanceQuerySBFCache
 					}
 				}
 
-				Concurrency::threadPool.CountCurrentThreadAsPaused();
```
calebwherry, line 70: This method and 'CountCurrentThreadAsResumed()' are no longer used anywhere but they are still in ThreadPool.h. Should they be removed?

src/Amalgam/evaluablenode/EvaluableNodeManagement.cpp
```diff
@@ -244,51 +241,52 @@ void EvaluableNodeManager::FreeAllNodes()
...
 		#ifdef MULTITHREAD_SUPPORT
-			//before releasing the lock, make sure it has an allocated type, otherwise it could get grabbed by another thread
+			//before releasing the lock, make sure the EvaluableNode is initialized, otherwise it could get grabbed by another thread
 			nodes[firstUnusedNodeIndex]->InitializeUnallocated();
 		#endif
 		}
 		else //allocate if nullptr
+		{
 			nodes[firstUnusedNodeIndex] = new EvaluableNode();
+		}
```
calebwherry, line 289: Extra braces needed?

src/Amalgam/evaluablenode/EvaluableNodeManagement.cpp
```diff
@@ -302,12 +300,14 @@ EvaluableNode *EvaluableNodeManager::AllocUninitializedNode()
...
 	#ifdef MULTITHREAD_SUPPORT
-		//before releasing the lock, make sure it has an allocated type, otherwise it could get grabbed by another thread
+		//before releasing the lock, make sure the EvaluableNode is initialized, otherwise it could get grabbed by another thread
 		nodes[firstUnusedNodeIndex]->InitializeUnallocated();
 	#endif	
 	}
 	else //allocate if nullptr
+	{
 		nodes[firstUnusedNodeIndex] = new EvaluableNode();
+	}
```
calebwherry, line 310: Same here, extra braces needed? Usually they are left off.

src/Amalgam/evaluablenode/EvaluableNodeManagement.cpp
```diff
@@ -335,8 +333,9 @@ void EvaluableNodeManager::FreeAllNodesExceptReferencedNodes()
 		auto &cur_node_ptr = nodes[first_unused_node_index_temp];
 
 		//if the node has been found on this iteration and set to the current iteration count, then move on
```
calebwherry: Since the logic changed, this comment thoulg probably be updated as well.

src/Amalgam/evaluablenode/EvaluableNodeManagement.cpp
```diff
@@ -736,6 +730,75 @@ void EvaluableNodeManager::NonCycleModifyLabelsForNodeTree(EvaluableNode *tree,
 	}
 }
 
+void EvaluableNodeManager::MarkAllReferencedNodesInUse(bool set_in_use)
+{
+#ifdef MULTITHREAD_SUPPORT
+	size_t reference_count = nodesCurrentlyReferenced.size();
+	//heuristic to ensure there's enough to do to warrant the overhead of using multiple threads
+	if(reference_count > 1 && firstUnusedNodeIndex / reference_count >= 2000)
```
calebwherry: How do you feel about lack of parens here? I'm with "operator precedence is obvious" but I would lean more towards parens.

src/Amalgam/KnnCache.h
howsohazard: Sure, sounds good, since the next PBI we discussed will revamp how they work anyway.

src/Amalgam/evaluablenode/EvaluableNodeManagement.cpp
howsohazard: They are because it matches the pattern of the first one... which needs to be in braces because it has a preprocessor, otherwise the code block looks quite different than the pattern.  Make sense, or do you have a better proposal?

src/Amalgam/evaluablenode/EvaluableNodeManagement.cpp
howsohazard: See previous comment

src/Amalgam/evaluablenode/EvaluableNodeManagement.cpp
calebwherry: Nah, I figured that was why, but wanted to make sure.

src/Amalgam/evaluablenode/EvaluableNodeManagement.cpp
howsohazard: Good catch, updated

src/Amalgam/evaluablenode/EvaluableNodeManagement.cpp
howsohazard: Updated

calebwherry (approved):

PR #69, opened 2024-02-01
19133: Improves multithread CPU utilization

howsohazard:


src/Amalgam/ThreadPool.h
```diff
@@ -11,26 +11,42 @@
...
-		return numActiveThreads;
+		ShutdownAllThreads();
 	}
 
+	//changes the maximum number of active threads
+	void SetMaxNumActiveThreads(int32_t max_num_active_threads);
+
 	//returns the current maximum number of threads that are available
-	inline size_t GetCurrentMaxNumThreads()
+	constexpr int32_t GetMaxNumActiveThreads()
```
calebwherry, line 41: Why `constexpr` here instead of `inline`? This value isn't known at compile time.

src/Amalgam/ThreadPool.h
```diff
@@ -11,26 +11,42 @@
...
 	{
-		std::unique_lock<std::mutex> lock(threadsMutex);
-		return threads.size();
+		return maxNumActiveThreads;
+	}
+
+	//returns the number of threads that are performing tasks
+	constexpr int32_t GetNumActiveThreads()
```
calebwherry, line 47: Same comment as `GetMaxNumActiveThreads`.

src/Amalgam/ThreadPool.h
```diff
@@ -196,26 +249,50 @@ class ThreadPool
...
+	//if positive, as threads become available they can decrement the value
+	//transition to reserved.  if negative, then reserved threads can increment
+	//the value to become available
+	int32_t numThreadsToTransitionToReserved;
+
+	//if true, then all threads should end work so they can be joined
+	bool shutdownThreads;
 
 	//id of the main thread
 	std::thread::id mainThreadId;
```
calebwherry, line 298: Can we remove this? I don't think it's useful and is actually misleading when loading Amalgam as a library (i.e., the thread that loads the library is not guaranteed to be a 'main thread' like in an app, most likely it will not be).

src/Amalgam/ThreadPool.h
```diff
@@ -196,26 +249,50 @@ class ThreadPool
 		return result;
 	}
 
-private:
+protected:
```
calebwherry, line 253: Reason to move from `private` to `protected` here? Looks like everything is internal and we don't inherit from this class.

src/Amalgam/ThreadPool.h
```diff
@@ -196,26 +249,50 @@ class ThreadPool
...
 	std::queue<std::function<void()>> taskQueue;
 
+	//the number of threads that can be active at any time
+	//the total number of threads is
+	//numActiveThreads + numReservedThreads + number of idle threads
+	int32_t maxNumActiveThreads;
+
 	//number of threads running
-	std::atomic<size_t> numActiveThreads;
+	//atomic so that it can be read dynamically without a lock
```
calebwherry: Do you mean that int32_t RWs are atomic so it can be accessed without a lock? If so, this comment probably isn't neccessary. Or did you mean for this to be a std::atomic?

src/Amalgam/ThreadPool.h
```diff
@@ -49,12 +65,48 @@ class ThreadPool
 	//returns true if there are threads currently idle
 	inline bool AreThreadsAvailable()
 	{
-		std::unique_lock<std::mutex> lock(taskQueueMutex);
-		return (taskQueue.size() + numActiveThreads < threads.size());
+		std::unique_lock<std::mutex> lock(threadsMutex);
+		//need to make sure there's at least one extra thread available to make sure that this batch of tasks can be run
+		// in case there are any interdependencies, in order to prevent deadlock
+		//need to take into account upcoming tasks, as they may consume threads
+		return ((numActiveThreads - numThreadsToTransitionToReserved) + 1 + static_cast<int32_t>(taskQueue.size()) <= maxNumActiveThreads);
```
calebwherry, line 72: This number is used at least 2 times and maybe 3, I think putting it in an inline function would be a good idea.

src/Amalgam/ThreadPool.h
```diff
@@ -49,12 +65,48 @@ class ThreadPool
...
+	//changes the current thread state from active to waiting
+	//the thread must currently be active
+	//this is intended to be called before waiting for other threads to complete their tasks
+	inline void ChangeCurrentThreadStateFromActiveToWaiting()
+	{
+		{
+			std::unique_lock<std::mutex> lock(threadsMutex);
+
+			//only add a new thread if no reserved and at capacity
+			if(numReservedThreads == 0 && (numActiveThreads - numThreadsToTransitionToReserved) + 1 == maxNumActiveThreads)
```
calebwherry, line 85: Why doesn't this calculation take into account `static_cast<int32_t>(taskQueue.size())` like the other 2 calculations do that are compared to `maxNumActiveThreads`?

src/Amalgam/ThreadPool.cpp
```diff
@@ -1,110 +1,143 @@
 //project headers:
 #include "ThreadPool.h"
+#include <iostream>
```
calebwherry, line 5: Add newline and a comment "//system headers:".

src/Amalgam/ThreadPool.cpp
```diff
@@ -1,110 +1,143 @@
...
+		lock.lock();
+
 		threads.clear();
 
 		//no longer shutting down, allow to build up threads
 		shutdownThreads = false;
-	}
 
-	size_t num_new_threads = new_max_num_threads - threads.size();
+		//reset other stats
```
calebwherry, line 41: Should `numThreadsToTransitionToReserved` also be reset here?

calebwherry: If so, having a `Reset()` method might be useful since it is duplicated in the constructor.

src/Amalgam/interpreter/Interpreter.h
```diff
@@ -548,9 +548,11 @@ class Interpreter
 		//ends concurrency from all interpreters and waits for them to finish
 		inline void EndConcurrency()
 		{
+			Concurrency::threadPool.ChangeCurrentThreadStateFromActiveToWaiting();
 			//make sure all futures return before moving on
```
calebwherry: I'd remove this comment, seems unneccesary.

src/Amalgam/evaluablenode/EvaluableNodeManagement.cpp
```diff
@@ -782,9 +790,13 @@ void EvaluableNodeManager::MarkAllReferencedNodesInUse(bool set_in_use)
 
 			enqueue_task_lock.Unlock();
 
+			Concurrency::threadPool.ChangeCurrentThreadStateFromActiveToWaiting();
+
 			for(auto& future : nodes_completed)
 				future.wait();
 
+			Concurrency::threadPool.ChangeCurrentThreadStateFromWaitingToActive();
```
calebwherry: Remove extra newlines to match style of other 2 places this pattern was added.

src/Amalgam/ThreadPool.cpp
```diff
@@ -1,110 +1,143 @@
...
+
+					//wait until either shutting down or a thread is requested to come out of reserved
+					waitForActivate.wait(lock,
+						[this] { return numThreadsToTransitionToReserved < 0 || shutdownThreads; });
+
+					//only can make it here if shutting down (otherwise taskQueue has something in it)
+					if(shutdownThreads)
+						return;
+
+					//coming out of reserved, go active unless no task
```
calebwherry: Is this comment accurate? Not sure about the "... unless no task" part is relevant here.

src/Amalgam/ThreadPool.h
howsohazard: Because it could be resolved during construction in certain circumstances for optimization, plus constexpr implies inline and const.  If you think it's misleading, I could change it to inline, but the usage is also elsewhere.

src/Amalgam/ThreadPool.h
howsohazard: See other comment.

src/Amalgam/ThreadPool.h
howsohazard: It's very much needed for multithreaded debugging.  I suppose it could in theory be removed for dll/so/dylib builds, but would need to have other preprocessors elsewhere for that.

src/Amalgam/ThreadPool.h
howsohazard: Philosophically, private means absolutely don't touch, protected means it could be touched by something that extends it.  I could imagine something extending it and wanting to access some of those variables.  It doesn't matter much though.  If you feel strongly, I could change it, but just figured protected is a more natural.

src/Amalgam/ThreadPool.h
howsohazard: outdated comment, fixed!

src/Amalgam/ThreadPool.h
howsohazard: It's not checking if there's capacity for a new task, it's checking if there will be insufficient reserved threads.

howsohazard: Updated the comment to reflect

src/Amalgam/ThreadPool.cpp
howsohazard: done

src/Amalgam/ThreadPool.cpp
howsohazard: outdated comment -- fixed

src/Amalgam/ThreadPool.h
calebwherry: Nope, that explanation is enough.

src/Amalgam/ThreadPool.h
calebwherry: Hmm, I'll follow up about this, interested why/how you use it in MT debugging.

src/Amalgam/ThreadPool.h
calebwherry: Sounds good to me.

calebwherry (approved):

PR #75, opened 2024-02-20
19185: Slightly improves CPU and memory utilization via urgent thread pool and minor memory reuse improvements

howsohazard:


src/Amalgam/Concurrency.h
```diff
@@ -82,6 +82,7 @@ namespace Concurrency
 
 #ifdef MULTITHREAD_SUPPORT
 	extern ThreadPool threadPool;
+	extern ThreadPool urgentThreadPool;
```
calebwherry: I think a comment here on the usage between the 2 would be a good idea. Something to give a at least some guidance on when/how to use one or the other.

src/Amalgam/interpreter/InterpreterOpcodesTransformations.cpp
```diff
@@ -643,7 +643,7 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_WEAVE(EvaluableNode *en, b
 	EvaluableNodeReference woven_list(evaluableNodeManager->AllocNode(ENT_LIST), true);
 
 	//just lists, interleave
-	if(function == nullptr)
+	if(EvaluableNode::IsNull(function))
```
calebwherry, line 646: Is this a general style we should move to for null checking ENs?

src/Amalgam/SeparableBoxFilterDataStore.h
```diff
@@ -144,28 +144,22 @@ class SeparableBoxFilterDataStore
...
+				columns_completed.emplace_back(
+					Concurrency::urgentThreadPool.EnqueueSingleTask([this, &entities, i]() { BuildLabel(i, entities); })
+				);
+			}
 
-				Concurrency::threadPool.ChangeCurrentThreadStateFromActiveToWaiting();
-				for(auto &future : columns_completed)
-					future.wait();
-				Concurrency::threadPool.ChangeCurrentThreadStateFromWaitingToActive();
+			Concurrency::urgentThreadPool.ChangeCurrentThreadStateFromActiveToWaiting();
```
calebwherry, line 157: Is the current thread guaranteed to be in the urgentThreadPool?

calebwherry: I suppose it doesn't actually matter if it is an urgent thread or not but still... seems like we are changing the urgentpool thread state but it could be from a non-urgentpool thread.

src/Amalgam/interpreter/InterpreterOpcodesTransformations.cpp
howsohazard: If you want to check for it being a null *value*, yes, this is what we should use.

calebwherry: Ah, got it, makes sense.

calebwherry (approved):

PR #82, opened 2024-02-25
19429: Fixes issue where zero weighted features sometimes apply feature attributes for incorrect features in queries

howsohazard:


src/Amalgam/entity/EntityQueryBuilder.h
```diff
@@ -429,6 +429,52 @@ namespace EntityQueryBuilder
...
+				{
+					//add label to the exist_condition
+					auto label_sid = cur_condition->positionLabels[i];
+					exist_condition->existLabels.push_back(label_sid);
+
+					//remove label
+					cur_condition->distParams.featureParams.erase(begin(cur_condition->distParams.featureParams) + i);
+					cur_condition->positionLabels.erase(begin(cur_condition->positionLabels) + i);
+
+					if(!DoesDistanceQueryUseEntitiesInsteadOfPosition(cur_condition->queryType))
```
howsoRes: should this condition be cached into a boolean outside the 'for' statement since it seems that it won't change per label? (and it'll be checked at least once since there is at least one unused feature)

howsoRes (approved):

PR #85, opened 2024-02-28
18116: Refactors generalized distances and separates general and repeated distance computation caches

howsohazard:


src/Amalgam/entity/EntityQueryBuilder.h
```diff
@@ -39,84 +39,84 @@ namespace EntityQueryBuilder
...
+						case ENBISI_nominal_numeric:			feature_type = GeneralizedDistanceEvaluator::FDT_NOMINAL_NUMERIC;			break;
+						case ENBISI_nominal_string:				feature_type = GeneralizedDistanceEvaluator::FDT_NOMINAL_STRING;				break;
+						case ENBISI_nominal_code:				feature_type = GeneralizedDistanceEvaluator::FDT_NOMINAL_CODE;				break;
+						case ENBISI_continuous_numeric:			feature_type = GeneralizedDistanceEvaluator::FDT_CONTINUOUS_NUMERIC;			break;
+						case ENBISI_continuous_numeric_cyclic:	feature_type = GeneralizedDistanceEvaluator::FDT_CONTINUOUS_NUMERIC_CYCLIC;	break;
+						case ENBISI_continuous_string:			feature_type = GeneralizedDistanceEvaluator::FDT_CONTINUOUS_STRING;			break;
+						case ENBISI_continuous_code:			feature_type = GeneralizedDistanceEvaluator::FDT_CONTINUOUS_CODE;			break;
+
+						default:								feature_type = GeneralizedDistanceEvaluator::FDT_CONTINUOUS_NUMERIC;			break;
 						}
```
calebwherry: The `break`s here don't show up as aligned in VSCode diff. It isn't clear if this is just a diff thing or the tabs or off in the formatting...

src/Amalgam/entity/EntityQueryBuilder.h
```diff
@@ -297,7 +296,6 @@ namespace EntityQueryBuilder
 		//else don't bother parsing this, it instead contains the cases to compute case conviction for
 
 		size_t num_elements = cur_condition->positionLabels.size();
-		auto &dist_params = cur_condition->distParams;
```
calebwherry, line 300: Doesn't removing this cause a bunch more derefs since you are using cur_condition->distParams a lot in changes below?

src/Amalgam/entity/EntityQueryCaches.cpp
```diff
@@ -259,15 +259,15 @@ void EntityQueryCaches::GetMatchingEntities(EntityQueryCondition *cond, BitArray
...
 
-			//only select cases that have all of the correct features
-			//but remove features that have 0 weight for better performance
+			//only keep entities that have all the correct features,
+			//but remove 0 weighted features for better performance
 			for(size_t i = 0; i < cond->positionLabels.size(); i++)
 			{
 				sbfds.IntersectEntitiesWithFeature(cond->positionLabels[i], matching_entities, true);
-				if(!cond->distParams.IsFeatureEnabled(i))
+				if(cond->distEvaluator.featureAttribs[i].weight == 0.0)
```
calebwherry, line 267: You removed the `IsFeatureEnabled` call here, is `== 0.0` the same thing or should both be checked?

src/Amalgam/entity/EntityQueryBuilder.h
howsohazard: from find-replace; fixed

src/Amalgam/entity/EntityQueryCaches.cpp
howsohazard: I removed IsFeatureEnabled entirely :-).  It just used a variation on this logic.

src/Amalgam/entity/EntityQueryBuilder.h
howsohazard: A few, but not that many.  I didn't think it was worth keeping.  Take a look and let me know if you think otherwise

howsohazard: Also, this isn't in an inner loop, so readability is slightly more important here.

calebwherry: Works for me, resolving.

calebwherry (dismissed):

src/Amalgam/interpreter/InterpreterOpcodesDataTypes.cpp
```diff
@@ -1029,32 +1029,12 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_ZIP_LABELS(EvaluableNode *
...
-		//if it has a label, it could have been a previous child node, therefore can't guarantee uniqueness
-		if(cur_value->GetNumLabels() > 0)
-			retval.unique = false;
+		if(retval_ocn[i] == nullptr)
+			retval_ocn[i] = evaluableNodeManager->AllocNode(ENT_NULL);
+		else if(!source.unique)
+			retval_ocn[i] = evaluableNodeManager->AllocNode(retval_ocn[i]);
 
-		cur_value->AppendLabelStringId(label_sid, true);
+		retval_ocn[i]->AppendLabelStringId(label_sid, true);
```
calebwherry: In the old code there are paths that set `reval.unique` to `false`. But that is not done in your updates. Is that because another change does that? Or it's not needed anymore? `retval` is a ref so changing it does change something... Not clear to me if that is still needed since it is still set in line 1026 above.

Comment on line 1012 seems to indicate that it should be set when needed. But I could be misunderstanding things here at a higher/lower level.

howsohazard: Improved!  Good call out, and I updated the comments as well.

calebwherry (dismissed):

calebwherry (approved):

PR #92, opened 2024-03-08
19551: Fixes an issue where tracefile label names are in quotes

cademack:
Running tracefiles with label names surrounded in quotes was not executing the labels correctly. 

My change here moves and renames `Platform_SplitArgString` to `StringManipulation::SplitArgString` (and moves it from PlatformSpecific.h/cpp to StringManipulation.h/cpp)

Additionally, a boolean flag is added to this method named `greedy`. As the name suggests, if `greedy` is True, then the method returns the full vector of args. If it is False, then a vector with only the first arg is returned and the argument string is modified to have that first word cut out.

Then, `RemoveFirstWord` is renamed to `RemoveFirstToken` and modified to use this new modified version of `SplitArgString` with the `greedy` flag set to False to remove the first word, then returns an empty string if no string is returned.

### DIFF
--- a/src/Amalgam/string/StringManipulation.cpp
+++ b/src/Amalgam/string/StringManipulation.cpp
@@ -58,7 +58,7 @@ std::string StringManipulation::RemoveFirstWord(std::string &str, bool strip_wor
 {
 	std::string first_token;
 
-	if(str.size() == 0)
+	if(str.empty())
 		return first_token;
 
 
@@ -108,7 +108,7 @@ std::string StringManipulation::RemoveFirstWord(std::string &str, bool strip_wor
 	}
 
 	//remove preceding spaces in str
-	if(str.size() > 0 && str[0] == ' ')
+	if(!str.empty() && str[0] == ' ')
 	{
 		size_t cur_pos = 1;
 		while(cur_pos < str.size() && str[cur_pos] == ' ')
--- a/src/Amalgam/string/StringManipulation.cpp
+++ b/src/Amalgam/string/StringManipulation.cpp
@@ -54,22 +54,15 @@ std::string StringManipulation::NumberToString(size_t value)
 	return std::string(&buffer[0]);
 }
 
-std::string StringManipulation::RemoveFirstWord(std::string &str)
+std::string StringManipulation::RemoveFirstToken(std::string &str)
 {
 	std::vector<std::string> arg;
-	arg = StringManipulation::SplitArgString(str, true);
+	arg = StringManipulation::SplitArgString(str, false);
 
-	if(arg.empty())
-	{
-		return "";
-	}
-	else
-	{
-		return arg[0];
-	}
+	return (arg.empty() ? "" : arg[0]);
 }
 
-std::vector<std::string> StringManipulation::SplitArgString(std::string &arg_string, bool non_greedy)
+std::vector<std::string> StringManipulation::SplitArgString(std::string &arg_string, bool greedy)
 {
 	std::vector<std::string> args;
 
@@ -122,7 +115,7 @@ std::vector<std::string> StringManipulation::SplitArgString(std::string &arg_str
 		}
 		args.push_back(cur_arg);
 
-		if(non_greedy)
+		if(!greedy)
 		{
 			arg_string = arg_string.substr(cur_pos);
 			return args;
--- a/src/Amalgam/string/StringManipulation.cpp
+++ b/src/Amalgam/string/StringManipulation.cpp
@@ -57,7 +57,6 @@ std::string StringManipulation::NumberToString(size_t value)
 std::string StringManipulation::RemoveFirstWord(std::string &str)
 {
 	std::vector<std::string> arg;
-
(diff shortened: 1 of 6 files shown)


howsoRes (dismissed):

src/Amalgam/string/StringManipulation.cpp
```diff
@@ -54,20 +54,70 @@ std::string StringManipulation::NumberToString(size_t value)
...
 }
 
-std::string StringManipulation::RemoveFirstWord(std::string &str)
+std::string StringManipulation::RemoveFirstWord(std::string &str, bool strip_word, char char_to_strip)
 {
 	std::string first_token;
-	size_t spacepos = str.find(' ');
-	if(spacepos == std::string::npos)
+
+	if(str.size() == 0)
```
calebwherry: I'd use `str.empty()`.

cademack: good idea, changed.

src/Amalgam/string/StringManipulation.cpp
```diff
@@ -54,20 +54,70 @@ std::string StringManipulation::NumberToString(size_t value)
...
+std::string StringManipulation::RemoveFirstWord(std::string &str, bool strip_word, char char_to_strip)
 {
 	std::string first_token;
-	size_t spacepos = str.find(' ');
-	if(spacepos == std::string::npos)
+
+	if(str.empty())
+		return first_token;
+
+
```
calebwherry, line 68: Remove extra empty line.

src/Amalgam/string/StringManipulation.cpp
```diff
@@ -56,19 +56,80 @@ std::string StringManipulation::NumberToString(size_t value)
...
-		first_token = str;
-		str = "";
+		return "";
 	}
 	else
 	{
-		first_token = str.substr(0, spacepos);
-		str = str.substr(spacepos + 1);
+		return arg[0];
 	}
```
calebwherry: Use a ternary operator here: `return (arg.empty() ? "" : arg[0])`

src/Amalgam/string/StringManipulation.h
```diff
@@ -13,9 +13,16 @@ namespace StringManipulation
 	std::string NumberToString(size_t value);
 
 	//removes the first word from str and return the removed word
+	//if strip_word true, remove char_to_strip from beginning and end of returned word
```
calebwherry: Remove this comment.

src/Amalgam/string/StringManipulation.h
```diff
@@ -13,9 +13,16 @@ namespace StringManipulation
...
 	std::string RemoveFirstWord(std::string &str);
```
calebwherry: I would rename this to `RemoveFirstToken` and update call sites.

src/Amalgam/string/StringManipulation.h
```diff
@@ -13,9 +13,16 @@ namespace StringManipulation
...
 
-	std::vector<std::string> Split(std::string &s, char delim);
+	//splits a string by given delimiter
+	std::vector<std::string> Split(std::string &s, char delim = ' ');
+
+	//separates the argument string in a cross-platform manner and returns an appropriate vector of strings
```
calebwherry: I would remove "cross-platform manner" here. I don't think it adds any information, anything outside of "PlatformSpecific" should be cross platform by definition.

src/Amalgam/string/StringManipulation.h
```diff
@@ -13,9 +13,16 @@ namespace StringManipulation
...
+	//if non_greedy is true, the returned vector only contains a single string, and arg_string is modified
+	//to only contain the portion of the string after the removed section
+	std::vector<std::string> SplitArgString(std::string &arg_string, bool non_greedy = false);
```
calebwherry: I think it makes more sense to make this `greedy` and default to true. I know originally said the other way but double negative is a bit ugly looking. Be sure to update the code too.

src/Amalgam/string/StringManipulation.h
```diff
@@ -13,9 +13,15 @@ namespace StringManipulation
 	std::string NumberToString(size_t value);
 
 	//removes the first word from str and return the removed word
```
calebwherry: Update comment to match new name.

src/Amalgam/string/StringManipulation.h
```diff
@@ -13,9 +13,15 @@ namespace StringManipulation
...
-	std::string RemoveFirstWord(std::string &str);
+	std::string RemoveFirstToken(std::string &str);
 
-	std::vector<std::string> Split(std::string &s, char delim);
+	//splits a string by given delimiter
+	std::vector<std::string> Split(std::string &s, char delim = ' ');
+
+	//separates the argument string and returns an appropriate vector of strings
+	//if non_greedy is true, the returned vector only contains a single string, and arg_string is modified
+	//to only contain the portion of the string after the removed section
```
calebwherry: Update comment to match new param name.

calebwherry (approved):

PR #97, opened 2024-03-18
19678: Changes caml format to automatically flatten and reconstruct entities, MAJOR

howsohazard:


docs/language.js
```diff
@@ -1364,7 +1364,7 @@ var data = [
...
... defaults to false, but if it is true, it will agressively escape filenames using only alphanumeric characters and the underscore, using underscore as an escape character.  If escape_contained_filenames is true, which is its default, it will also escape contained entity filenames.  If file_type is specified and not null, it will use the file_type specified instead of the extension of the file_path.  File formats supported are amlg, json, yaml, csv, and caml; anything not in this list will be loaded as a binary string.  Note that loading from a non-'.amlg' extension will only ever provide lists, assocs, numbers, and strings.  If params is specified, it is an assoc that contains key-value pairs describing the format.  The key \"sort_keys\" can be used to specify a boolean value, if true, then it will sort the keys, otherwise the default behavior is to emit the keys based on memory layout.  The key \"include_rand_seeds\" can be used when storing caml files to indicate whether random seeds will be stored, which defaults to true.  The key \"parallel_create\" can be used when storing caml files to indicate whether creating entities will be performed in parallel when the caml is loaded.",
```
calebwherry: What is the default for 'parallel_create'?

src/Amalgam/out.txt
```diff
@@ -3782,7 +3771,11 @@ contained entity in quackers2 loaded back from file: (list "buklulu is good.amlg
 store to .amlg:
 2
 2
-Compression difference: [(null)]
+Compression difference: [(declare
+	(assoc _ (null) new_entity (null))
+	(clone_entities _ new_entity)
+)
+]
```
calebwherry, line 3778: What caused this diff?

howsohazard: This is what it should have been all along -- showing that they are identical because the code just clones it.

src/Amalgam/AssetManager.cpp
```diff
@@ -192,6 +175,21 @@ Entity *AssetManager::LoadEntityFromResourcePath(std::string &resource_path, std
...
+
+		EvaluableNodeReference args = EvaluableNodeReference(new_entity->evaluableNodeManager.AllocNode(ENT_ASSOC), true);
+		args->SetMappedChildNode(ENBISI_create_new_entity, new_entity->evaluableNodeManager.AllocNode(ENT_FALSE));
+		auto call_stack = Interpreter::ConvertArgsToCallStack(args, new_entity->evaluableNodeManager);
+
+		ExecutionCycleCount max_num_steps = 0, num_steps_executed = 0;
+		size_t max_num_nodes = 0, num_nodes_allocated = 0;
+		new_entity->Execute(max_num_steps, num_steps_executed, max_num_nodes, num_nodes_allocated, StringInternPool::NOT_A_STRING_ID, call_stack);
+		return new_entity;
+	}
```
calebwherry, line 191: Doesn't this execute with root permissions?

howsohazard: No -- look at EntityExternalInterface::LoadEntity -- it sets root permissions after load.

howsohazard: We can consider removing permissions there and adding new API methods to enable different permissions -- though that's out of scope for this task.

calebwherry: Second question: this gets executed for every loaded caml regardless if user wants it or not. How does that affect behavior? What's the before/after expectation of existing code?

calebwherry: Another question here: what about all the other logic after this block? It's not clear to me that all that getting skipped is fine now...

howsohazard: Entities stored as camls are code to reconstruct the entity.  What happened before (currently) is that this needs to be loaded and then executed.  This simplifies it and achieves the intent of being able to reconstruct all contained entities.  The rest of the method isn't needed because all of the contained entities are constructed from this code.  But if it's not all in one file, then each needs to be loaded separately (the rest of the method).

calebwherry (approved):

PR #115, opened 2024-04-09
19916: Improves approximate deviation accuracy that fixes issue where nearest neighbors would be missed with large deviations

howsohazard:


src/Amalgam/evaluablenode/EvaluableNodeTreeManipulation.h
```diff
@@ -55,7 +55,7 @@ inline double NumberCommonality(double difference, double a, double b)
 {
 	double max_abs = std::max(std::fabs(a), std::fabs(b));
 	//since this is called frequently in comparing and merging, and perfect accuracy isn't required, just use fast version
```
howsoRes: the comment is now out of date.  please update to explain the reason for the cast to float

howsoRes (approved):

PR #116, opened 2024-04-10
19919: Changes debugging line numbers to be 1-based instead of 0-based

howsohazard:


fulpm (dismissed):

fulpm (commented):
Do the warning outputs in this file also need to be incremented by one? See lines 452 and 599. Seems like this only affects the comments output for the debugger.

fulpm (approved):

PR #155, opened 2024-06-19
20490: Unifies .nas, and .nan under null keyword, MAJOR

howsohazard:
.nas and .nan have been removed and have been replaced simply by (null)


src/Amalgam/SeparableBoxFilterDataStore.cpp
```diff
@@ -7,6 +7,8 @@ thread_local
 #endif
 SeparableBoxFilterDataStore::SBFDSParametersAndBuffers SeparableBoxFilterDataStore::parametersAndBuffers;
 
+//TODO 20490: check for NAS and remove as appropriate
```
cademack: Is this TODO still necessary?

cademack (dismissed):
All looks fine enough

cademack (dismissed):

docs/language.js
```diff
@@ -1549,7 +1549,7 @@ var data = [
 		"parameter" : "query_mode string label_name [string weight_label_name] [bool numeric]",
 		"output" : "query",
 		"new value" : "new",
-		"description" : "When used as a query argument, finds the statistical mode of label_name for numerical data.  If weight_label_name is specified, it will find the weighted mode.  If numeric is true, its default, then it will treat all values as numeric, otherwise it will treat them all as strings.  If numeric and no numeric mode exists, it will return .nan, but if string and no string mode exists, it will return null.",
+		"description" : "When used as a query argument, finds the statistical mode of label_name for numerical data.  If weight_label_name is specified, it will find the weighted mode.  If numeric is true, its default, then it will treat all values as numeric, otherwise it will treat them all as strings.  If numeric and no numeric mode exists, it will return (null), but if string and no string mode exists, it will return null.",
```
howsoRes, line 1552: please search language.js for all instances of nan and nans (not case sensitive) and update accordingly as well. it's in many other places besides this one.

src/Amalgam/SBFDSColumnData.h
```diff
@@ -221,28 +218,9 @@ class SBFDSColumnData
 			{
 				double old_number_value = GetResolvedValue(old_value_type, old_value).number;
 				double new_number_value = GetResolvedValue(new_value_type, new_value).number;
-				if(EqualIncludingNaN(old_number_value, new_number_value))
+				if(old_number_value == new_number_value)
```
howsoRes, line 221: you seem to have removed all usages of EqualIncludingNaN() except from line 1066 of SBFDS.cpp.  since that's its only use, maybe copy-paste that one liner directly into that cpp and get rid of that function?

howsoRes (approved):

src/Amalgam/SBFDSColumnData.h
howsohazard: it's used elsewhere, like the double comparator

PR #156, opened 2024-06-24
20655: Initializes EvaluableNode to type ENT_NULL if double value is NaN

cademack:
Note, this merges into branch 20490, not main

### DIFF
--- a/src/Amalgam/evaluablenode/EvaluableNode.h
+++ b/src/Amalgam/evaluablenode/EvaluableNode.h
@@ -90,16 +90,19 @@ public:
 
 	constexpr void InitializeType(double number_value)
 	{
-		if(std::isnan(number_value))
+		if(FastIsNaN(number_value))
+		{
 			type = ENT_NULL;
+			value.ConstructOrderedChildNodes();
+		}
 		else
 		{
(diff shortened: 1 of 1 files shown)


src/Amalgam/evaluablenode/EvaluableNode.h
```diff
@@ -90,11 +90,16 @@ class EvaluableNode
 
 	constexpr void InitializeType(double number_value)
 	{
-		type = ENT_NUMBER;
-		attributes.allAttributes = 0;
-		attributes.individualAttribs.isIdempotent = true;
-		value.numberValueContainer.labelStringID = StringInternPool::NOT_A_STRING_ID;
-		value.numberValueContainer.numberValue = number_value;
+		if(std::isnan(number_value))
```
howsohazard: FastIsNaN

howsohazard: and should set all of the same values as below

howsohazard: actually, put this:
value.ConstructOrderedChildNodes();

howsohazard (approved):

PR #158, opened 2024-06-25
20668: Fixes `(null)`'s returned by `(indices)` to be typed as null

cademack:
The diff here is rendered strangely.

### DIFF
--- a/src/Amalgam/evaluablenode/EvaluableNode.cpp
+++ b/src/Amalgam/evaluablenode/EvaluableNode.cpp
@@ -659,23 +659,7 @@ void EvaluableNode::SetStringID(StringInternPool::StringID id)
 {
 	if(id == StringInternPool::NOT_A_STRING_ID)
 	{
-		if(DoesEvaluableNodeTypeUseStringData(GetType()))
-		{
-			if(!HasExtendedValue())
-			{
-				StringInternPool::StringID cur_id = value.stringValueContainer.stringID;
-				if(id != cur_id)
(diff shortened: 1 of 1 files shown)


src/Amalgam/evaluablenode/EvaluableNode.cpp
```diff
@@ -742,19 +742,24 @@ StringInternPool::StringID EvaluableNode::GetAndClearStringIDWithReference()
 
 void EvaluableNode::SetStringIDWithReferenceHandoff(StringInternPool::StringID id)
 {
-	if(DoesEvaluableNodeTypeUseStringData(GetType()))
+	if(id == StringInternPool::NOT_A_STRING_ID)
```
howsohazard, line 752: two notes here:
1) stylistically, if you use {}s in one branch of an if-else, use them for all
2) there's gonna need to be more logic here:
 a) if the current ID is a valid string ID, you have to destroy the reference (this requires both locations based on whether it has extended value)
 b) ordered child nodes needs to be initialized for null

howsohazard: Oh, also will need to change SetStringID too (the one without reference handoff)

cademack: Added the reference destroying logic to the NOT_A_STRING path, practically copied from the other branch, but without the assignments to the instances values. I figure they're not needed for a ENT_NULL node

howsohazard (approved):

PR #175, opened 2024-07-16
20879: Adds opcode depth performance constraints to many call opcodes, minor performance improvement, MINOR

howsohazard:


src/Amalgam/interpreter/Interpreter.h
```diff
@@ -786,44 +866,39 @@ class Interpreter
 		return true;
 	}
 
-	//recalculates curNumExecutionNodes
-	__forceinline void UpdateCurNumExecutionNodes()
+	//if true, no limit on how much memory can utilize
+	constexpr bool ConstrainedAllocatedNodes()
 	{
-		curNumExecutionNodes = curNumExecutionNodesAllocatedToEntities + evaluableNodeManager->GetNumberOfUsedNodes();
+		return (performanceConstraints != nullptr && performanceConstraints->ConstrainedAllocatedNodes());
```
howsoRes, line 872: shouldn't this also return true if performanceConsntrains are undefined (nullptr) as well?

howsoRes (approved):

PR #188, opened 2024-07-22
20999: Updated WASM test packages for security

lancegliser:
- Added additional local development setup

### DIFF
--- a/test/wasm_test/jest/global.setup.js
+++ b/test/wasm_test/jest/global.setup.js
@@ -2,5 +2,4 @@ import { config } from "dotenv";
 
 export default async () => {
   config();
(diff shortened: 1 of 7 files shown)


test/wasm_test/jest/global.setup.js
```diff
@@ -0,0 +1,5 @@
+import { config } from "dotenv";
+
+export default async () => {
+  config();
```
lancegliser, line 4: Note that `config` by default pulls `.env`. If it fails, it doesn't explode. It returns an object of which contains an error. The upstream pipelines should remain unaffected for the definition of `AMALGAM_WASM_DIR`.

howsohazard (approved):

PR #206, opened 2024-08-01
21093: Fixes issues related to storing and loading data structures with cyclic references

howsohazard:


src/Amalgam/Opcodes.h
```diff
@@ -309,7 +309,7 @@ enum OrderedChildNodeType
 	OCNT_POSITION
 };
 
-//Returns the type of structure that the ordered child nodes have for a given type
+//Returns the type of structure that the ordered child nodes have for a given t
```
cademack: is there a reason not to lowercase this "Returns" like the rest?

src/Amalgam/interpreter/InterpreterOpcodesBase.cpp
```diff
@@ -1269,7 +1269,7 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_TARGET(EvaluableNode *en,
 		double value = InterpretNodeIntoNumberValue(ocn[0]);
 		if(value >= 0)
 			depth = static_cast<size_t>(value);
-		else
+		else if(!FastIsNaN(value)) //null should leave depth as 0
```
cademack: probably still want an else branch, maybe if value is negative?

src/Amalgam/Parser.cpp
```diff
@@ -1070,39 +1096,52 @@ EvaluableNode *Parser::GetNodeFromRelativeCodePath(EvaluableNode *path)
...
-		size_t steps_up = 1;
+		size_t target_steps_up = 1;
 		if(path->GetOrderedChildNodes().size() > 0)
-			steps_up = static_cast<size_t>(EvaluableNode::ToNumber(path->GetOrderedChildNodes()[0]));
+		{
+			double step_value = EvaluableNode::ToNumber(path->GetOrderedChildNodes()[0]);
+			if(step_value >= 1)
+				target_steps_up = static_cast<size_t>(step_value);
+			else
+				return nullptr;
```
cademack, line 1134: In a similar spot you had a branch that set the step value to 0 if it was (null), should this spot do that as well?

howsoRes (approved):

PR #212, opened 2024-08-05
21133: Adds optional constraints to entity creation, fixes some bugs and resource leaks around failed entity creations, MINOR

howsohazard:


src/Amalgam/out.txt
apbassett, line 1: Is this file meant to be committed?

apbassett (approved):

src/Amalgam/out.txt
howsohazard: yes, it is how we do unit tests, by comparing the results from previous runs

PR #218, opened 2024-08-09
21187: Fixes bugs around constraining entity creation and mutating code

howsohazard:


src/Amalgam/evaluablenode/EvaluableNodeTreeManipulation.cpp
```diff
@@ -1505,7 +1505,7 @@ void MutateImmediateNode(EvaluableNode *n, RandomStream &rs, std::vector<std::st
 EvaluableNode *EvaluableNodeTreeManipulation::MutateNode(EvaluableNode *n, MutationParameters &mp)
 {
 	if(n == nullptr)
-		return nullptr;
+		n = mp.enm->AllocNode(ENT_NULL);
```
apbassett, line 1508: This is more for my own curiosity: why this change?

apbassett (approved):

src/Amalgam/evaluablenode/EvaluableNodeTreeManipulation.cpp
howsohazard: So that it can mutate a null as opposed to only keeping it the same

PR #220, opened 2024-08-12
21206: Fixes issue where unparsed code (e.g., camls, etc.) that had null as a key to an assoc would be emitted without the key

howsohazard:


src/Amalgam/evaluablenode/EvaluableNode.h
```diff
@@ -19,6 +19,7 @@
 #include <string>
 #include <vector>
 
+#define AMALGAM_FAST_MEMORY_INTEGRITY
```
jdbeel: Should this be here? If I'm reading the comment correctly, wouldn't this enable the fast memory integrity checks all the time?

jdbeel (approved):

cademack (approved):

apbassett (approved):

dmaze (approved):

fulpm (approved):

src/Amalgam/evaluablenode/EvaluableNode.h
howsohazard: I accidentally left that in and removed it, but not fast enough before you made your comment ;-)

PR #234, opened 2024-08-21
Cycle breaking

Paul-Licameli:
Untangle the include dependency graph of Amalgam, to produce a good summary overview diagram of its parts.

The "before" picture of dependencies, with a big strongly connected component of 25 compilation units:

graph dot

The picture after the cleanup:

graph dot

Now no two files outside 3rdparty are in an include dependency cycle, and the folder structure of files is also reorganized.  Boxes in the diagram correspond to folders.  It is not immediately evident from the picture but the quotient graph of include dependencies by folders is also acyclic.


howsohazard:
Many related cleanup changes have been made, making this branch obsolete

PR #266, opened 2024-09-13
21528: Adds debug target for WASM builds

apbassett:
WASM builds will now have two targets, standard release single-threaded and a debug single-threaded, the latter postfixed with `-debug`.

As a result of this, the standard `wasm64-debug` build presets have been removed as they are redundant.


lancegliser (dismissed):

build/cmake/custom_add_target.cmake
```diff
@@ -248,6 +251,11 @@ function(add_compiled_target)
         target_compile_options(${TARGET_NAME} PUBLIC -O1)
     endif()
 
+    if(IS_WASM AND USE_DEBUG)
```
howsohazard: Do we need a IS_WASM AND USE_DEBUG, or can we just have a single USE_DEBUG that applies to everything?

apbassett: Good catch, otherwise USE_DEBUG would be useless on non-WASM builds (which wasn't my intention!)

CMakeLists.txt
```diff
@@ -313,6 +313,15 @@ if(IS_LINUX AND IS_AMD64)
         SOURCE ${AMALGAM_LIB_ONLY_SOURCE} ${COMMON_SOURCE_THREADS} IDE_FOLDER "OtherBuildTargets")
 endif()
 
+# WASM single-threaded debug target (amalgam-st-debug)
+if(IS_WASM)
```
fulpm, line 317: Should we remove the wasm64-debug target? Looks like this will always build the debug variant along side the release variant. Might be cleaner to only build debug when its the target?

lancegliser: This was an ask from me. Can't be sure when any given release is going to need to have some debugging investigation. Publishing debug and production side by side to use as required. Also a help as we move the wasm side to open source in case we get third parties.

> Might be cleaner to only build debug when its the target?

I'd support granular build targets being available, long as the end release has them all.

apbassett: Yeah, I was initially debating a refactor to the CI/CD build script to simply execute release + debug builds for WASM (instead of shoehorning in a WASM debug target). The main draw towards the current solution for me was that it'd plop both WASMs in the same build artifact with different naming conventions (-st and -st-debug), so that would save a bit of a lift in the CI/CD. But perhaps that was the lazy way. @fulpm if you're fine with this and want me to remove the "proper" wasm-debug target, I will, but if you think it'd be best to simply use that and remove this shoehorned one, that's fine too. Just let me know.

fulpm: That is fine, lets just remove the `-debug` build presets since they wont be needed.

fulpm (dismissed):

fulpm (dismissed):

fulpm (approved):

PR #269, opened 2024-09-18
21566: Adds stack_distance and no_child_nodes parameters to (opcode_stack) and copy parameter to (previous_result), MINOR

cademack:
Adds two parameters to (opcode_stack):
- stack_distance, when specified only the specific opcode at the depth into the stack is returned
- no_child_nodes, when true only the root node(s) are copied and returned, otherwise (default) the root node(s) are deep copied and returned

Adds one parameter to (previous_result):
- copy, when true will copy the previous result node and return it, otherwise the previous result is returned directly and consumed.

Also includes a small bugfix for the (keep) opcode in the case that an index is not valid for an ordered container.


src/Amalgam/interpreter/InterpreterOpcodesBase.cpp
```diff
@@ -1358,19 +1358,94 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_PREVIOUS_RESULT(EvaluableN
...
+		double value = InterpretNodeIntoNumberValue(ocn[0]);
+		if(!FastIsNaN(value))
+		{
+			has_valid_depth = true;
+			depth = static_cast<int64_t>(value);
+		}
+	}
+	
+	bool no_child_nodes = false;
+	if (ocn.size() > 1)
```
howsohazard: no space after if

src/Amalgam/interpreter/InterpreterOpcodesBase.cpp
```diff
@@ -1358,19 +1358,94 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_PREVIOUS_RESULT(EvaluableN
...
+		no_child_nodes = InterpretNodeIntoBoolValue(ocn[1], false);
+
+	if(!has_valid_depth)
+	{
+		//return the whole opcode stack
+		//can create this node on the stack because will be making a copy
+		EvaluableNodeReference stack_top_holder(evaluableNodeManager->AllocNode(ENT_LIST), true);
```
howsohazard: Use the previous way of using the local stack for stack_top_holder when making a deep alloc copy, but use this method for if it's not a deep copy of the whole thing

howsohazard (dismissed):

howsohazard (approved):

PR #284, opened 2024-10-09
21803: Fixes bug with zip opcode where it would improperly type upgrade any nonimmediate value to a list for values, also corrects comments in SBFDS

howsohazard:


cademack (approved):

src/Amalgam/interpreter/InterpreterOpcodesTransformations.cpp
```diff
@@ -1639,6 +1639,7 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_ZIP(EvaluableNode *en, boo
 	if(value_list != nullptr)
 		result.UpdatePropertiesBasedOnAttachedNode(value_list, true);
 
+	bool value_list_is_a_list = (value_list != nullptr && value_list->GetType() == ENT_LIST);
```
naricc, line 1642: So basically, we are changing the check to see if its type is ENT_LIST, which I guess is not the same as checking with IsOrderedArray?

PR #298, opened 2024-10-28
21358: Reduces memory when loading entities with transactional flag

howsohazard:


howsoRes (approved):

src/Amalgam/AssetManager.cpp
```diff
@@ -241,6 +322,20 @@ Entity *AssetManager::LoadEntityFromResource(AssetParameters &asset_params, bool
 	Entity *new_entity = new Entity();
 	new_entity->SetRandomState(default_random_seed, true);
 
+	if(asset_params.executeOnLoad && asset_params.transactional)
+	{
+		if(!LoadResourceViaTransactionalExecution(asset_params, new_entity, calling_interpreter, status))
+		{
+			delete new_entity;
```
naricc, line 329: I'm not sure if this is a path worth optimizing, but there is an opportunity to avoid allocation here if we push allocation of new_entity into LoadResouceViaTransactionalExecution, and it just doesn't allocate it sometimes. Rather than allocated and then deleting.

howsohazard: Yeah, I considered that, but the fact that you need an EvaluableNodeManager to store the data makes it a lot clunkier without one in the first place, and the extra allocation only happens unnecessarily when you try to load something and it fails (error case).  So I lean towards leaving as-is unless you disagree.

naricc: I guess it would make the code too awkward for the benefit.

PR #305, opened 2024-11-18
22228: Reduces stack memory usage

howsohazard:


howsoRes (approved):

naricc:
What is the actual memory gain here? Just that we end up with less "extra" memory in the string because of the use of push_back instead of append?

PR #310, opened 2024-11-25
22173: Improves performance by implementing thread local allocations

naricc:
This is an improvement to performance of memroy management in amalgam. It changes allocation of EvaluableNodes so that we first try to allocate from a thread local buffer; only if that buffer is empty do we try to allocate from the EvaluableNodeManager pool. When eagerly freeing nodes, we also put them back in the thread local buffer.


src/Amalgam/evaluablenode/EvaluableNodeManagement.cpp
```diff
@@ -6,6 +6,7 @@
 #include <string>
 #include <vector>
 #include <utility>
+#include <iostream>
```
howsohazard: Leftover include

src/Amalgam/Parser.cpp
```diff
@@ -199,6 +199,8 @@ EvaluableNode *Parser::GetCodeForPathToSharedNodeFromParentAToParentB(UnparseDat
 	if(shared_node == nullptr || a_parent == nullptr || b_parent == nullptr)
 		return nullptr;
 
+	//TODO 22156: allocate from enm in a way that does not invalidate tlab (but does not increase the size of EvaluableNodeManager)
```
howsohazard: The urgency or importance of this is probably not high -- don't need to keep a TODO around in it isn't important

src/Amalgam/evaluablenode/EvaluableNodeManagement.cpp
```diff
@@ -59,63 +60,79 @@ EvaluableNode *EvaluableNodeManager::AllocNode(EvaluableNode *original, Evaluabl
...
+		// parent
+		node->InitializeType(ENT_LIST);
+		std::vector<EvaluableNode *> *ocn_ptr = &node->GetOrderedChildNodesReference();
+		std::swap(*ocn_buffer, *ocn_ptr);
+	}
+	else
+	{
+		// child node; initialize it and add it to the list items
+		std::vector<EvaluableNode *> *ocn_ptr = &parent->GetOrderedChildNodesReference();
+		(*ocn_ptr)[node_index-1] = node;
```
howsohazard: Simplify via:
auto &ocn = parent->GetOrderedChildNodesReference();
ocn[node_index-1] = node;

howsohazard: the first one above is to make the swap more clear, but it's possible that could be simplified as well

src/Amalgam/evaluablenode/EvaluableNodeManagement.cpp
```diff
@@ -59,63 +60,79 @@ EvaluableNode *EvaluableNodeManager::AllocNode(EvaluableNode *original, Evaluabl
...
 	{
+		EvaluableNode *newNode = nullptr;
+
+		while((newNode = GetNextNodeFromTLab()) && num_allocated < num_to_alloc)
 		{
-		#ifdef MULTITHREAD_SUPPORT
-			//attempt to allocate as many as possible using an atomic without write locking
-			Concurrency::ReadLock lock(managerAttributesMutex);
-		#endif
+			if(parent == nullptr)
```
howsohazard, line 103: Style is if there is only a single line for an if, then don't use {}'s

src/Amalgam/evaluablenode/EvaluableNodeManagement.cpp
```diff
@@ -59,63 +60,79 @@ EvaluableNode *EvaluableNodeManager::AllocNode(EvaluableNode *original, Evaluabl
...
-						ocn_ptr = &retval->GetOrderedChildNodesReference();
-						std::swap(ocn_buffer, *ocn_ptr);
-
-						//advance type to child node type
-						cur_type = child_node_type;
-					}
-					else //set the appropriate child node
-					{
-						(*ocn_ptr)[num_allocated - 1] = nodes[allocated_index];
+						nodes[allocated_index] = new EvaluableNode(ENT_DEALLOCATED);
```
howsohazard, line 128: Looks like the {}'s around this can be removed too.

src/Amalgam/evaluablenode/EvaluableNodeManagement.cpp
```diff
@@ -125,34 +142,40 @@ EvaluableNode *EvaluableNodeManager::AllocListNodeWithOrderedChildNodes(Evaluabl
 				}
 			}
 
-			//if have allocated enough, just return
-			if(num_allocated == num_to_alloc)
-				return retval;
-
-			num_total_nodes_needed = firstUnusedNodeIndex + (num_to_alloc - num_allocated);
+			if( num_added_to_tlab + num_allocated >= num_to_alloc)
```
howsohazard, line 142: put comment above, condense to something like "if enough nodes in tlab, use them in the next loop", remove {}'s

src/Amalgam/evaluablenode/EvaluableNodeManagement.h
```diff
@@ -1105,6 +1084,61 @@ class EvaluableNodeManager
 	//only allocated if needed
 	std::unique_ptr<NodesReferenced> nodesCurrentlyReferenced;
 
+public:
+
+	static void ClearThreadLocalAllocationBuffer()
```
howsohazard: Toss an inline on this.  Is the public needed above?

src/Amalgam/evaluablenode/EvaluableNodeManagement.h
```diff
@@ -1105,6 +1084,61 @@ class EvaluableNodeManager
...
+			threadLocalAllocationBuffer.clear();
+			lastEvaluableNodeManager = this;
+		}
+
+		threadLocalAllocationBuffer.push_back(en);
+	}
+
+	static const int tlabSize = 20;
+
+private:
```
howsohazard: Does this need to be private?

src/Amalgam/evaluablenode/EvaluableNodeManagement.h
```diff
@@ -1105,6 +1084,61 @@ class EvaluableNodeManager
...
+protected: 
 	//extra space to allocate when allocating
 	static const double allocExpansionFactor;
+
+#if defined(MULTITHREAD_SUPPORT) || defined(MULTITHREAD_INTERFACE)
+	thread_local
+#endif
+	static inline EvaluableNodeManager *lastEvaluableNodeManager;
+
+	inline EvaluableNode *GetNextNodeFromTLab()
```
howsohazard, line 1112: These methods need at least a brief comment.  Also, I don't think TLab is defined clearly anywhere, so it would be good to spell it out in the comments for these two methods and the thread_local variable

src/Amalgam/evaluablenode/EvaluableNodeManagement.h
naricc: It does need to be public since Interpreter calls it, but  I reorganized it so it so it can be with the other public members.

src/Amalgam/evaluablenode/EvaluableNodeManagement.h
naricc: Made it private.

howsohazard (approved):

PR #315, opened 2024-12-02
21364: Improves entity persistence: works as single transactional files as well as performance improvements, MINOR

howsohazard:


src/Amalgam/AssetManager.cpp
```diff
@@ -111,65 +117,162 @@ void AssetManager::AssetParameters::UpdateResources()
...
 
-EvaluableNodeReference AssetManager::LoadResource(AssetParameters &asset_params, EvaluableNodeManager *enm,
+//returns a pair of success followed by version number if a version can be found in an Amalgam metadata file
+static std::pair<bool, std::string> FindVersionStringInAmlgMetadata(std::ifstream &file)
+{
+	//read 200 bytes and null terminate
+	std::array<char, 201> buffer;
+	file.read(buffer.data(), 200);
+	buffer[200] = '\0';
+	std::string str(buffer.data(), 200);
```
dmaze: Is this sound if the file is shorter than 200 bytes?  In here (and probably the preceding line) do you need to use the return value from `file.read()` as the length/offset?

src/Amalgam/AssetManager.cpp
```diff
@@ -111,65 +117,162 @@ void AssetManager::AssetParameters::UpdateResources()
...
+}
+
+//returns a pair of success followed by version number if a version can be found in an Amalgam execute on load file
+std::pair<bool, std::string> FindVersionStringInAmlgExecOnLoad(std::ifstream &file)
+{
+	//read 200 bytes and null terminate
+	std::array<char, 201> buffer;
+	file.read(buffer.data(), 200);
+	buffer[200] = '\0';
+	std::string str(buffer.data(), 200);
```
dmaze: Should this code be split out into a helper function?  (I think the byte-level char array manipulation should be sketchy in modern C++ but can't immediately find an alternative, and it's easier to justify in comments just once.)

src/Amalgam/AssetManager.cpp
```diff
@@ -111,65 +117,162 @@ void AssetManager::AssetParameters::UpdateResources()
...
+	{
+		std::ifstream f(path, std::fstream::binary | std::fstream::in);
+		if(!f.good())
+			return std::make_tuple("Cannot open file", "", false);
+
+		return std::make_tuple("", "", true);
+	}
+}
+
+EvaluableNodeReference AssetManager::LoadResource(AssetParameters *asset_params, EvaluableNodeManager *enm,
```
dmaze, line 220: Should this be `AssetParameters &asset_params` (a reference and not a pointer) so that you don't have to change all of the uses of it?  (Do you ever reassign which `AssetParameters` it is?). Should it be a `const &`?  (Does the code ever change the parameters, or just read them?)

src/Amalgam/AssetManager.cpp
```diff
@@ -401,30 +553,30 @@ Entity *AssetManager::LoadEntityFromResource(AssetParameters &asset_params, bool
...
 
 	if(persistent)
-		SetEntityPersistence(new_entity, &asset_params);
+		SetEntityPersistence(new_entity, asset_params);
 
 	//load contained entities
 
 	//iterate over all files in directory
-	std::string contained_entities_directory = asset_params.resourceBasePath + "/";
+	std::string contained_entities_directory = asset_params->resourceBasePath + "/";
```
dmaze, line 561: If we're on C++17 or later, can we use [`std::filesystem::path`](https://en.cppreference.com/w/cpp/filesystem/path) here?

src/Amalgam/BinaryPacking.h
```diff
@@ -1,24 +1,109 @@
...
 
-//Appends the offset index oi to BinaryData
-void UnparseIndexToCompactIndexAndAppend(BinaryData &bd_out, OffsetIndex oi);
+//Huffman Encoding implementation for compressing and decompressing data
+template <typename value_type>
+class HuffmanTree
+{
+public:
+	constexpr HuffmanTree(value_type value, size_t value_frequency, size_t node_index,
+		HuffmanTree<value_type> *left = nullptr, HuffmanTree<value_type> *right = nullptr)
```
dmaze, line 17: Does using `std::unique_ptr<T>` make sense here?  Of note, that has the property that, if the unique pointer goes out of scope holding a value, it automatically deletes the value it owns (and you could remove the destructor).

howsohazard: Ideally yes, but I spent about an hour trying to get the compiler to be happy with unique_ptr here, but the downstream uses are extremely subtle and complex and then reasoned the extra complexity and fragility probably isn't worthwhile.  Switching over to shared/unique ptrs may be another hundred or more lines of highly specific copy/move/etc. constructors and/or accompanying documentation on exactly how they need to be used OR lots of unnecessary memory copies.

src/Amalgam/AssetManager.cpp
howsohazard: Yes, but doing so would require rewriting some of the file handling parts of PlatformSpecific that were written before C++17 was around.  Something we should do someday, but that's a larger effort.

src/Amalgam/AssetManager.cpp
howsohazard: If only you saw how many times I switch back and forth between passing as a reference and as a pointer in this work :-).  Many of the other methods do use shared pointers.  I could make two of the methods LoadResource and StoreResource work on AssetParams as a reference, but they'd be inconsistent with the rest that deal with pointers or shared pointers.  Would changing just those two be preferable?

src/Amalgam/AssetManager.cpp
howsohazard: Tightened the reading so that spurious old memory for implementations with loose interpretations of the C++ standards will still do the right thing.

src/Amalgam/AssetManager.cpp
howsohazard: It's in two spots really close to each other, each of which needs its own regular expression.  I could merge the methods, but don't know if that's better.

naricc (approved):

src/Amalgam/AssetManager.cpp
dmaze: (Doesn't need action now.)

src/Amalgam/AssetManager.cpp
dmaze: (Doesn't need action now.)

dmaze (approved):
I've run this branch separately on a couple of tests and it seems to work as intended.  I don't quite have the depth of knowledge on the Amalgam core to do a solid code review, but I can at least say it looks reasonable and it works.

PR #316, opened 2024-12-02
22199 Implements index_min and index_max op codes, MINOR

naricc:
This adds index_min and index_max op codes to amalgam, as well as adding calls to full_test.amlg

### DIFF
--- a/docs/language.js
+++ b/docs/language.js
@@ -508,16 +508,16 @@ var data = [
 	},
 	{
 		"parameter" : "index_max [number x1] [number x2] [number x3] ... [number xN] | (assoc ...) | (list ...)",
-		"output" : "*",
+		"output" : "[*]",
 		"concurrency" : true,
-		"description" : "the index of of the argument with the greatest value, or the key associated with the greatest value in an assoc, or the index of the greatest value in a list",
+		"description" : "a list of the indices of the argument with the greatest value, or the keys associated with the greatest value in an assoc, or the indices of the greatest value in a list",
 		"example" : "(print (index_max 0.5 1 7 9 -5))\n(print (index_max (list 0.5 1 7 9 -5) ))\n(print (index_max (assoc 1 0.5 2 1 3 7)))"
 	},
 	{
 		"parameter" : "index_min [number x1] [number x2] [number x3] ... [number xN] | (assoc ...) | (list ...)",
-		"output" : "*",
+		"output" : "[*]",
 		"concurrency" : true,
-		"description" : "the index of of the argument with the least value, or the key associated with the least value in an assoc, or the index of the least value in a list",
+		"description" : "a list of the indices of of the arguments with the least value, or the keys associated with the least value in an assoc, or the indices of the least value in a list",
 		"example" : "(print (index_min 0.5 1 7 9 -5))\n(print (index_min (list 0.5 1 7 9 -5) ))\n(print (index_min (assoc 1 0.5 2 1 3 7)))"
 	},
 	{
--- a/docs/language.js
+++ b/docs/language.js
@@ -506,7 +506,20 @@ var data = [
 		"description" : "minimum of all of the numbers",
 		"example" : "(print (min 0.5 1 7 9 -5))"
 	},
-
+	{
+		"parameter" : "index_max list|assoc",
+		"output" : "index|key",
+		"concurrency" : true,
+		"description" : "the list index or assoc key mapped to the maximum value",
+		"example" : "(print (index_max 0.5 1 7 9 -5))"
+	},
+	{
+		"parameter" : "index_min list|assoc",
+		"output" : "index|key",
+		"concurrency" : true,
+		"description" : "the list index or assoc key mapped to the minimum value",
+		"example" : "(print (index_min 0.5 1 7 9 -5))"
+	},
 	{
 		"parameter" : "dot_product list|assoc x1 list|assoc x2",
 		"output" : "number",
--- a/docs/language.js
+++ b/docs/language.js
@@ -507,17 +507,17 @@ var data = [
 		"example" : "(print (min 0.5 1 7 9 -5))"
 	},
 	{
-		"parameter" : "index_max *",
+		"parameter" : "index_max [number x1] [number x2] [number x3] ... [number xN] | (assoc ...) | (list ...)",
 		"output" : "*",
 		"concurrency" : true,
-		"description" : "the list index or assoc key mapped to the maximum value",
+		"description" : "the index of of the argument with the greatest value, or the key associated with the greatest value in an assoc, or the index of the greatest value in a list",
 		"example" : "(print (index_max 0.5 1 7 9 -5))\n(print (index_max (list 0.5 1 7 9 -5) ))\n(print (index_max (assoc 1 0.5 2 1 3 7)))"
 	},
 	{
-		"parameter" : "index_min *",
+		"parameter" : "index_min [number x1] [number x2] [number x3] ... [number xN] | (assoc ...) | (list ...)",
 		"output" : "*",
 		"concurrency" : true,
-		"description" : "the list index or assoc key mapped to the minimum value",
+		"description" : "the index of of the argument with the least value, or the key associated with the least value in an assoc, or the index of the least value in a list",
 		"example" : "(print (index_min 0.5 1 7 9 -5))\n(print (index_min (list 0.5 1 7 9 -5) ))\n(print (index_min (assoc 1 0.5 2 1 3 7)))"
 	},
 	{
--- a/docs/language.js
+++ b/docs/language.js
@@ -507,15 +507,15 @@ var data = [
 		"example" : "(print (min 0.5 1 7 9 -5))"
 	},
 	{
-		"parameter" : "index_max list|assoc",
-		"output" : "index|key",
+		"parameter" : "index_max *",
+		"output" : "*",
 		"concurrency" : true,
 		"description" : "the list index or assoc key mapped to the maximum value",
 		"example" : "(print (index_max 0.5 1 7 9 -5))"
 	},
 	{
-		"parameter" : "index_min list|assoc",
-		"output" : "index|key",
+		"parameter" : "index_min *",
+		"output" : "*",
 		"concurrency" : true,
 		"description" : "the list index or assoc key mapped to the minimum value",
 		"example" : "(print (index_min 0.5 1 7 9 -5))"
--- a/docs/language.js
+++ b/docs/language.js
@@ -511,14 +511,14 @@ var data = [
 		"output" : "*",
 		"concurrency" : true,
 		"description" : "the list index or assoc key mapped to the maximum value",
-		"example" : "(print (index_max 0.5 1 7 9 -5))"
+		"example" : "(print (index_max 0.5 1 7 9 -5))\n(print (index_max (list 0.5 1 7 9 -5) ))\n(print (index_max (assoc 1 0.5 2 1 3 7)))"
 	},
 	{
 		"parameter" : "index_min *",
 		"output" : "*",
 		"concurrency" : true,
 		"description" : "the list index or assoc key mapped to the minimum value",
-		"example" : "(print (index_min 0.5 1 7 9 -5))"
+		"example" : "(print (index_min 0.5 1 7 9 -5))\n(print (index_min (list 0.5 1 7 9 -5) ))\n(print (index_min (assoc 1 0.5 2 1 3 7)))"
 	},
 	{
 		"parameter" : "dot_product list|assoc x1 list|assoc x2",
--- a/docs/language.js
+++ b/docs/language.js
@@ -510,14 +510,14 @@ var data = [
 		"parameter" : "index_max [number x1] [number x2] [number x3] ... [number xN] | (assoc ...) | (list ...)",
 		"output" : "[*]",
 		"concurrency" : true,
-		"description" : "a list of the indices of the argument with the greatest value, or the keys associated with the greatest value in an assoc, or the indices of the greatest value in a list",
+		"description" : "If given multiple arguments, returns a list of the indices of the arguments with the maximum value. If given a single argument that is an assoc returns the a list of keys associated with the maximum values. If given a single argument that is a list, returns a list of list indices with the maximum value.",
 		"example" : "(print (index_max 0.5 1 7 9 -5))\n(print (index_max (list 0.5 1 7 9 -5) ))\n(print (index_max (assoc 1 0.5 2 1 3 7)))"
 	},
 	{
 		"parameter" : "index_min [number x1] [number x2] [number x3] ... [number xN] | (assoc ...) | (list ...)",
 		"output" : "[*]",
 		"concurrency" : true,
-		"description" : "a list of the indices of of the arguments with the least value, or the keys associated with the least value in an assoc, or the indices of the least value in a list",
+		"description" : "If given multiple arguments, returns a list of the indices of the arguments with the minimum value. If given a single argument that is an assoc returns the a list of keys associated with the minimum values. If given a single argument that is a list, returns a list of list indices with the minimum value.",
 		"example" : "(print (index_min 0.5 1 7 9 -5))\n(print (index_min (list 0.5 1 7 9 -5) ))\n(print (index_min (assoc 1 0.5 2 1 3 7)))"
 	},
 	{
--- a/src/Amalgam/Opcodes.cpp
+++ b/src/Amalgam/Opcodes.cpp
@@ -112,6 +112,8 @@ void StringInternPool::InitializeStaticStrings()
 	EmplaceNodeTypeString(ENT_ABS, "abs");
 	EmplaceNodeTypeString(ENT_MAX, "max");
 	EmplaceNodeTypeString(ENT_MIN, "min");
+	EmplaceNodeTypeString(ENT_INDEX_MAX, "index_max");
+	EmplaceNodeTypeString(ENT_INDEX_MIN, "index_min");
 	EmplaceNodeTypeString(ENT_GENERALIZED_DISTANCE, "generalized_distance");
 	EmplaceNodeTypeString(ENT_DOT_PRODUCT, "dot_product");
 	EmplaceNodeTypeString(ENT_ENTROPY, "entropy");
--- a/src/Amalgam/Opcodes.h
+++ b/src/Amalgam/Opcodes.h
@@ -466,7 +466,7 @@ constexpr OpcodeNewValueReturnType GetOpcodeNewValueReturnType(EvaluableNodeType
 	case ENT_ASIN:	case ENT_COS:	case ENT_ACOS:	case ENT_TAN:	case ENT_ATAN:
 	case ENT_SINH:	case ENT_ASINH:	case ENT_COSH:	case ENT_ACOSH:	case ENT_TANH:
 	case ENT_ATANH:	case ENT_ERF:	case ENT_TGAMMA:	case ENT_LGAMMA:	case ENT_SQRT:
-	case ENT_POW:	case ENT_ABS:	case ENT_MAX:	case ENT_MIN:	case ENT_DOT_PRODUCT:
+	case ENT_POW:	case ENT_ABS:	case ENT_MAX:	case ENT_INDEX_MAX: case ENT_INDEX_MIN: case ENT_MIN: case ENT_DOT_PRODUCT:
 	case ENT_GENERALIZED_DISTANCE:	case ENT_ENTROPY:
 	case ENT_SIZE:
 	case ENT_REWRITE:
--- a/src/Amalgam/Opcodes.h
+++ b/src/Amalgam/Opcodes.h
@@ -91,6 +91,8 @@ enum EvaluableNodeType : uint8_t
 	ENT_ABS,
 	ENT_MAX,
 	ENT_MIN,
+	ENT_INDEX_MAX,
+	ENT_INDEX_MIN,
 	ENT_DOT_PRODUCT,
 	ENT_GENERALIZED_DISTANCE,
 	ENT_ENTROPY,
@@ -313,7 +315,7 @@ constexpr OrderedChildNodeType GetOpcodeOrderedChildNodeType(EvaluableNodeType t
 	case ENT_PARALLEL:
 	case ENT_ADD:
 	case ENT_MULTIPLY:
-	case ENT_MAX:					case ENT_MIN:
+	case ENT_MAX:					case ENT_MIN:				case ENT_INDEX_MAX:	case ENT_INDEX_MIN:
 	case ENT_AND:					case ENT_OR:				case ENT_XOR:
 	case ENT_EQUAL:					case ENT_NEQUAL:
 	case ENT_NULL:
--- a/src/Amalgam/amlg_code/full_test.amlg
+++ b/src/Amalgam/amlg_code/full_test.amlg
@@ -399,6 +399,7 @@
  (print (index_max .5 -12 3 5 7))
  (print (index_max (list 1 1 3 2 1 3)))
  (print (index_max (null) 34  -66))
+ (print (index_max "NaN" "NaN" "NaN") "\n")
  (print (index_max (assoc  1 2 "tomato" 4444 3 5) ))
 
  (print "--index_min--\n")
--- a/src/Amalgam/amlg_code/full_test.amlg
+++ b/src/Amalgam/amlg_code/full_test.amlg
@@ -395,6 +395,14 @@
  (print (min 0.5 1 7 9 -5) "\n")
  (print (min (null) 4 8) "\n")
 
+ (print "--index_max--")
+ (print (index_max .5 -12 3 5 7))
+ (print (index_max (null) 34  -66))
+
+ (print "--index_min--")
+ (print (index_min .5 -12 3 5 7))
+ (print (index_min (null) 34  -66))
+
  (print "--dot_product--\n")
  (print (dot_product (list 0.5 0.25 0.25) (list 4 8 8)) "\n")
  (print (dot_product (associate "a" 0.5 "b" 0.25 "c" 0.25) (associate "a" 4 "b" 8 "c" 8)) "\n")
--- a/src/Amalgam/amlg_code/full_test.amlg
+++ b/src/Amalgam/amlg_code/full_test.amlg
@@ -396,16 +396,16 @@
  (print (min (null) 4 8) "\n")
 
  (print "--index_max--\n")
- (print (index_max .5 -12 3 5 7) "\n")
- (print (index_max (list 1 2 3)) "\n")
- (print (index_max (null) 34  -66) "\n")q
- (print (index_max (assoc  1 2 "tomato" 4444 3 5) ) "\n")
+ (print (index_max .5 -12 3 5 7))
+ (print (index_max (list 1 1 3 2 1 3)))
+ (print (index_max (null) 34  -66))
+ (print (index_max (assoc  1 2 "tomato" 4444 3 5) ))
 
  (print "--index_min--\n")
- (print (index_min .5 -12 3 5 7) "\n")
- (print (index_min (list 1 2 3)) "\n")
- (print (index_min (null) 34  -66) "\n")
- (print (index_min (assoc  1 2 "tomato" 4444 3 5) ) "\n")
+ (print (index_min .5 -12 3 5 7))
+ (print (index_min (list 1 1 3 2 1 3)))
+ (print (index_min (null) 34  -66))
+ (print (index_min (assoc  1 2 "tomato" 4444 3 5) ))
 
  (print "--dot_product--\n")
  (print (dot_product (list 0.5 0.25 0.25) (list 4 8 8)) "\n")
--- a/src/Amalgam/amlg_code/full_test.amlg
+++ b/src/Amalgam/amlg_code/full_test.amlg
@@ -395,18 +395,19 @@
  (print (min 0.5 1 7 9 -5) "\n")
  (print (min (null) 4 8) "\n")
 
- (print "--index_max--")
- (print (index_max .5 -12 3 5 7))
- (print (index_max (null) 34  -66))
+ (print "--index_max--\n")
+ (print (index_max .5 -12 3 5 7) "\n")
+ (print (index_max (null) 34  -66) "\n")
+ (print (index_max (assoc  1 2 "tomato" 4444 3 5) ) "\n")
 
- (print "--index_min--")
- (print (index_min .5 -12 3 5 7))
- (print (index_min (null) 34  -66))
+ (print "--index_min--\n")
+ (print (index_min .5 -12 3 5 7) "\n")
+ (print (index_min (null) 34  -66) "\n")
+ (print (index_min (assoc  1 2 "tomato" 4444 3 5) ) "\n")
 
  (print "--dot_product--\n")
  (print (dot_product (list 0.5 0.25 0.25) (list 4 8 8)) "\n")
  (print (dot_product (associate "a" 0.5 "b" 0.25 "c" 0.25) (associate "a" 4 "b" 8 "c" 8)) "\n")
-
  (print (dot_product (associate 0 0.5 1 0.25 2 0.25) (list 4 8 8)) "\n")
 
  (print "--generalized_distance--\n")
--- a/src/Amalgam/amlg_code/full_test.amlg
+++ b/src/Amalgam/amlg_code/full_test.amlg
@@ -399,7 +399,7 @@
  (print (index_max .5 -12 3 5 7))
  (print (index_max (list 1 1 3 2 1 3)))
  (print (index_max (null) 34  -66))
- (print (index_max "NaN" "NaN" "NaN") "\n")
+ (print (index_max (null) (null) (null)) "\n")
  (print (index_max (assoc  1 2 "tomato" 4444 3 5) ))
 
  (print "--index_min--\n")
--- a/src/Amalgam/amlg_code/full_test.amlg
+++ b/src/Amalgam/amlg_code/full_test.amlg
@@ -397,11 +397,13 @@
 
  (print "--index_max--\n")
  (print (index_max .5 -12 3 5 7) "\n")
+ (print (index_max (list 1 2 3)) "\n")
  (print (index_max (null) 34  -66) "\n")q
  (print (index_max (assoc  1 2 "tomato" 4444 3 5) ) "\n")
 
  (print "--index_min--\n")
  (print (index_min .5 -12 3 5 7) "\n")
+ (print (index_min (list 1 2 3)) "\n")
  (print (index_min (null) 34  -66) "\n")
  (print (index_min (assoc  1 2 "tomato" 4444 3 5) ) "\n")
 
--- a/src/Amalgam/amlg_code/full_test.amlg
+++ b/src/Amalgam/amlg_code/full_test.amlg
@@ -397,7 +397,7 @@
 
  (print "--index_max--\n")
  (print (index_max .5 -12 3 5 7) "\n")
- (print (index_max (null) 34  -66) "\n")
+ (print (index_max (null) 34  -66) "\n")q
  (print (index_max (assoc  1 2 "tomato" 4444 3 5) ) "\n")
 
  (print "--index_min--\n")
--- a/src/Amalgam/evaluablenode/EvaluableNode.cpp
+++ b/src/Amalgam/evaluablenode/EvaluableNode.cpp
@@ -1274,7 +1274,7 @@ void EvaluableNode::AppendOrderedChildNodes(const std::vector<EvaluableNode *> &
 	if(!IsOrderedArray())
 		return;
 
-	auto &ocn = GetOrderedChildNodesReference ();
+	auto &ocn = GetOrderedChildNodesReference();
 	ocn.insert(end(ocn), begin(ocn_to_append), end(ocn_to_append));
 
 	//if cycles, propagate upward
--- a/src/Amalgam/evaluablenode/EvaluableNode.cpp
+++ b/src/Amalgam/evaluablenode/EvaluableNode.cpp
@@ -1274,7 +1274,7 @@ void EvaluableNode::AppendOrderedChildNodes(const std::vector<EvaluableNode *> &
 	if(!IsOrderedArray())
 		return;
 
-	auto &ocn = GetOrderedChildNodesReference();
+	auto &ocn = GetOrderedChildNodesReference ();
 	ocn.insert(end(ocn), begin(ocn_to_append), end(ocn_to_append));
 
 	//if cycles, propagate upward
@@ -1282,7 +1282,7 @@ void EvaluableNode::AppendOrderedChildNodes(const std::vector<EvaluableNode *> &
 	{
 		if(cn != nullptr && cn->GetNeedCycleCheck())
 		{
-			SetNeedCycleCheck(true);
+		SetNeedCycleCheck(true);
 			break;
 		}
 	}
--- a/src/Amalgam/evaluablenode/EvaluableNode.cpp
+++ b/src/Amalgam/evaluablenode/EvaluableNode.cpp
@@ -1282,7 +1282,7 @@ void EvaluableNode::AppendOrderedChildNodes(const std::vector<EvaluableNode *> &
 	{
 		if(cn != nullptr && cn->GetNeedCycleCheck())
 		{
-		SetNeedCycleCheck(true);
+			SetNeedCycleCheck(true);
 			break;
 		}
 	}
--- a/src/Amalgam/evaluablenode/EvaluableNode.h
+++ b/src/Amalgam/evaluablenode/EvaluableNode.h
@@ -880,18 +880,19 @@ public:
 		debugWatch.erase(en);
 	}
 
-	//returns true if the EvaluableNode is in the debug watch
+	// returns true if the EvaluableNode is in the debug watch
 	static inline void AssertIfInDebugWatch(EvaluableNode *en)
 	{
 	#if defined(MULTITHREAD_SUPPORT) || defined(MULTITHREAD_INTERFACE)
 		Concurrency::SingleLock lock(debugWatchMutex);
 	#endif
-		if(debugWatch.find(en) != end(debugWatch)) {
+		if(debugWatch.find(en) != end(debugWatch))
+		{
 			assert(false);
 		}
 	}
 
-protected:
+  protected:
 
 	//align to the nearest 2-bytes to minimize alignment issues but reduce the overall memory footprint
 	// while maintaining some alignment
--- a/src/Amalgam/evaluablenode/EvaluableNode.h
+++ b/src/Amalgam/evaluablenode/EvaluableNode.h
@@ -1163,7 +1163,7 @@ class EvaluableNodeImmediateValueWithType
 {
 public:
 	constexpr EvaluableNodeImmediateValueWithType()
-		: nodeType(ENIVT_NULL)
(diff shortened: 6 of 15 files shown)


src/Amalgam/interpreter/InterpreterOpcodesMath.cpp
```diff
@@ -805,6 +806,128 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_MIN(EvaluableNode *en, boo
...
+static bool lessThanCompareFunc(double x, double y)
+{
+	return x < y;
+}
+
+static bool greaterThanCompareFunc(double x, double y)
+{
+	return x > y;
+}
+
```
howsohazard: Just use std::greater and std::less instead of defining own.

src/Amalgam/interpreter/InterpreterOpcodesMath.cpp
```diff
@@ -805,6 +806,128 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_MIN(EvaluableNode *en, boo
...
+EvaluableNodeReference Interpreter::InterpretNode_ENT_INDEX_MAX(EvaluableNode *en, bool immediate_result)
+{
+	auto &ocn = en->GetOrderedChildNodes();
+
+	if(ocn.size() == 0)
+		return EvaluableNodeReference::Null();
+
+	
+	if(ocn[0]->GetType() == ENT_ASSOC)
```
howsohazard: Need to check to make sure ocn[0] is not a nullptr

src/Amalgam/interpreter/InterpreterOpcodesMath.cpp
howsohazard: Will need to update function signatures that accept these as well.

src/Amalgam/interpreter/InterpreterOpcodesMath.cpp
```diff
@@ -805,6 +806,128 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_MIN(EvaluableNode *en, boo
...
+	bool value_found = false;
+	double result_value = compare_limit;
+
+#ifdef MULTITHREAD_SUPPORT
+	std::vector<EvaluableNodeReference> interpreted_nodes;
+	if(InterpretEvaluableNodesConcurrently(en, ocn, interpreted_nodes, true))
+	{
+
+		size_t max_index = 0;
+		for (size_t i = 0; i < interpreted_nodes.size(); i++)
```
howsohazard: no space between built-in opcodes and open paren -- should be "for(".

src/Amalgam/interpreter/InterpreterOpcodesMath.cpp
```diff
@@ -805,6 +806,128 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_MIN(EvaluableNode *en, boo
...
+			return AllocReturn(static_cast<double>(max_index), immediate_result);
+
+		return EvaluableNodeReference::Null();
+	}
+#endif
+
+	auto node_stack = CreateOpcodeStackStateSaver();
+
+	size_t max_index = 0;
+	// for(auto &cn : ocn)
```
howsohazard: left-in commented out code

src/Amalgam/interpreter/InterpreterOpcodesMath.cpp
```diff
@@ -805,6 +806,128 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_MIN(EvaluableNode *en, boo
...
+		return EvaluableNodeReference::Null();
+
+	bool value_found = false;
+	double result_value = compare_limit;
+
+#ifdef MULTITHREAD_SUPPORT
+	std::vector<EvaluableNodeReference> interpreted_nodes;
+	if(InterpretEvaluableNodesConcurrently(en, ocn, interpreted_nodes, true))
+	{
+
```
howsohazard, line 838: probably drop the extra newline here

src/Amalgam/interpreter/InterpreterOpcodesMath.cpp
```diff
@@ -805,6 +806,128 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_MIN(EvaluableNode *en, boo
...
+	}
+#endif
+
+	auto node_stack = CreateOpcodeStackStateSaver();
+
+	size_t max_index = 0;
+	// for(auto &cn : ocn)
+	for(size_t i = 0; i < ocn.size(); i++)
+	{
+
```
howsohazard, line 838: drop extra newline here too

src/Amalgam/interpreter/InterpreterOpcodesMath.cpp
```diff
@@ -805,6 +806,128 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_MIN(EvaluableNode *en, boo
...
+		if(compareFunc(cur_value, candidate_value))
+		{
+			candidate_value = cur_value;
+			max_key = cur_key;
+			value_found = true;
+		}
+	}
+
+	if(value_found)
+		return  AllocReturn(max_key, false);
```
howsohazard: extra space between return & AllocReturn

src/Amalgam/interpreter/InterpreterOpcodesMath.cpp
```diff
@@ -805,6 +806,128 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_MIN(EvaluableNode *en, boo
...
+
+EvaluableNodeReference Interpreter::InterpretNode_IndexMinMaxAssoc(EvaluableNode *assoc, bool (*compareFunc)(double, double), double compare_limit, bool immediate_result )
+{
+	EvaluableNode::AssocType  mapped_child_nodes = assoc->GetMappedChildNodesReference();
+	double candidate_value = compare_limit;
+	bool value_found = false;
+
+	StringInternPool::StringID max_key;
+
+	for (auto [cur_key, cur_child]: mapped_child_nodes)
```
howsohazard: extra space between for and paren

src/Amalgam/interpreter/InterpreterOpcodesMath.cpp
```diff
@@ -805,6 +806,128 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_MIN(EvaluableNode *en, boo
...
+
+		if(compareFunc(cur_value, candidate_value))
+		{
+			candidate_value = cur_value;
+			max_key = cur_key;
+			value_found = true;
+		}
+	}
+
+	if(value_found)
```
howsohazard: will need to call Parser::ParseFromKeyString on max_key rather than AllocReturn in case a key is a data type other than a string.

src/Amalgam/interpreter/InterpreterOpcodesMath.cpp
```diff
@@ -7,6 +7,7 @@
 #include <cstdlib>
 
 #include <utility>
+#include <EvaluableNode.h>
```
howsohazard: Follow header include style found in other files

src/Amalgam/amlg_code/full_test.amlg
```diff
@@ -395,10 +395,19 @@
...
 
+ (print "--index_max--\n")
+ (print (index_max .5 -12 3 5 7) "\n")
+ (print (index_max (null) 34  -66) "\n")
+ (print (index_max (assoc  1 2 "tomato" 4444 3 5) ) "\n")
+
+ (print "--index_min--\n")
+ (print (index_min .5 -12 3 5 7) "\n")
+ (print (index_min (null) 34  -66) "\n")
+ (print (index_min (assoc  1 2 "tomato" 4444 3 5) ) "\n")
```
howsohazard: should update out.txt with the PR (though that is often the source of merge conflicts, so wait until the end with everything else updated)

src/Amalgam/Opcodes.cpp
```diff
@@ -113,6 +113,8 @@ void StringInternPool::InitializeStaticStrings()
 	EmplaceNodeTypeString(ENT_ABS, "abs");
 	EmplaceNodeTypeString(ENT_MAX, "max");
 	EmplaceNodeTypeString(ENT_MIN, "min");
+	EmplaceNodeTypeString(ENT_INDEX_MAX, "index_max");
+	EmplaceNodeTypeString(ENT_INDEX_MIN, "index_min");
```
howsohazard, line 116: Need to update language.js in docs with these new opcodes

docs/language.js
```diff
@@ -397,7 +397,20 @@ var data = [
 		"description" : "minimum of all of the numbers",
 		"example" : "(print (min 0.5 1 7 9 -5))"
 	},
-
+	{
+		"parameter" : "index_max list|assoc",
+		"output" : "index|key",
```
howsohazard: Should be *, meaning code/anything

docs/language.js
```diff
@@ -397,7 +397,20 @@ var data = [
...
+		"concurrency" : true,
+		"description" : "the list index or assoc key mapped to the maximum value",
+		"example" : "(print (index_max 0.5 1 7 9 -5))"
+	},
+	{
+		"parameter" : "index_min list|assoc",
+		"output" : "index|key",
```
howsohazard: * as well

src/Amalgam/interpreter/Interpreter.h
```diff
@@ -1309,6 +1311,91 @@ class Interpreter
 	//a stack (list) of the current nodes being executed
 	std::vector<EvaluableNode *> *opcodeStackNodes;
 
+	template <typename Compare>
+	EvaluableNodeReference InterpretNode_IndexMinMaxAssoc(EvaluableNode *assoc, Compare compare, double compare_limit, bool immediate_result )
+	{
```
howsohazard, line 1282: extra space before )

src/Amalgam/interpreter/Interpreter.h
```diff
@@ -1309,6 +1311,91 @@ class Interpreter
...
+		EvaluableNode::AssocType  mapped_child_nodes = assoc->GetMappedChildNodesReference();
+		double candidate_value = compare_limit;
+		bool value_found = false;
+
+		StringInternPool::StringID max_key;
+
+		for (auto [cur_key, cur_child]: mapped_child_nodes)
```
howsohazard: remove space after for

src/Amalgam/interpreter/Interpreter.h
```diff
@@ -1309,6 +1311,91 @@ class Interpreter
...
+			return EvaluableNodeReference::Null();
+
+		bool value_found = false;
+		double result_value = compare_limit;
+
+	#ifdef MULTITHREAD_SUPPORT
+		std::vector<EvaluableNodeReference> interpreted_nodes;
+		if(InterpretEvaluableNodesConcurrently(en, ocn, interpreted_nodes, true))
+		{
+
```
howsohazard, line 1336: remove extra newline

src/Amalgam/interpreter/Interpreter.h
```diff
@@ -1309,6 +1311,91 @@ class Interpreter
...
+			return EvaluableNodeReference::Null();
+		}
+	#endif
+
+		auto node_stack = CreateOpcodeStackStateSaver();
+
+		size_t max_index = 0;
+		for(size_t i = 0; i < ocn.size(); i++)
+		{
+
```
howsohazard, line 1343: remove extra newline

src/Amalgam/interpreter/InterpreterOpcodesMath.cpp
howsohazard: Still need to check ocn[0] isn't nullptr before calling GetType.  Also, remove extra newline above

src/Amalgam/interpreter/InterpreterOpcodesMath.cpp
```diff
@@ -805,6 +806,33 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_MIN(EvaluableNode *en, boo
...
+		return InterpretNode_IndexMinMaxAssoc(ocn[0], std::greater(), -std::numeric_limits<double>::infinity(), immediate_result);
+	else
+		return InterpretNode_IndexMinMaxList(en, std::greater(), -std::numeric_limits<double>::infinity(), immediate_result);
+}
+
+EvaluableNodeReference Interpreter::InterpretNode_ENT_INDEX_MIN(EvaluableNode *en, bool immediate_result)
+{
+	auto &ocn = en->GetOrderedChildNodes();
+
+	if(ocn.size() == 0 || ocn[0] == nullptr)
```
howsohazard: What should the behavior be if you have (index_min (null) 1 2 3) ?  Currently it will return null.  Should detail this in the documentation if this is what you're aiming for and make it consistent with index_max.

naricc: Changed the behavior so that would now it treats nulls as NaN's (like max/min does).

src/Amalgam/interpreter/InterpreterOpcodesMath.cpp
```diff
@@ -805,6 +806,32 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_MIN(EvaluableNode *en, boo
...
+{
+	auto &ocn = en->GetOrderedChildNodes();
+
+	if(ocn.size() == 0)
+		return EvaluableNodeReference::Null();
+
+	if(ocn[0] != nullptr && ocn[0]->GetType() == ENT_ASSOC)
+		return InterpretNode_IndexMinMaxAssoc(ocn[0], std::greater(), -std::numeric_limits<double>::infinity(), immediate_result);
+	else
+		return InterpretNode_IndexMinMaxList(en, std::greater(), -std::numeric_limits<double>::infinity(), immediate_result);
```
howsohazard: This method doesn't handle nullptr either, so could still crash if nullptr.  Same with ENT_INDEX_MIN.  Also recommend using standard naming for the methods that are not opcodes, i.e. GetIndexMinMaxFromList / Assoc without InterpretNode_ in front.  Could make a static method to keep it local to the file.

naricc: I don't think I can easily make this static because  using std::greater/less obligates me to make it a template. Unless there is some way I am not thinking of?

I added null checks to InterpretNode_IndexMinMaxList though.

src/Amalgam/interpreter/InterpreterOpcodesMath.cpp
```diff
@@ -1,12 +1,13 @@
 //project headers:
 #include "Interpreter.h"
-
+#include "EvaluableNode.h"
 #include "EntityQueryBuilder.h"
 
 //system headers:
 #include <cstdlib>
```
howsohazard, line 7: System headers should be included in alphabetical order with no spaces

src/Amalgam/interpreter/Interpreter.h
```diff
@@ -1309,6 +1311,89 @@ class Interpreter
...
+		if(value_found)
+			return  Parser::ParseFromKeyString(max_key->string, evaluableNodeManager);
+
+		return EvaluableNodeReference::Null();
+	}
+
+	template <typename Compare>
+	EvaluableNodeReference GetIndexMinMaxFromList(EvaluableNode *en, Compare compare, double compare_limit, bool immediate_result)
+	{
+		auto &ocn = en->GetOrderedChildNodes();
```
howsohazard: en may be nullptr which can cause a segfault here (e.g., in the test if it had been (+ (null)).  Further, it would either segfault, or the ocn == nullptr below would never be true.  Recommend an early out when checking if size is zero in the opcodes themselves, and assume that en is valid going into each of these methods.  Further, for consistency, I would imagine that (index_max (list 1 2 3)) should evaluate to 2 (third index), but I don't think that's the case, and the opcode seems inconsistent.  The documentation in language.js should also be updated and clarified.

docs/language.js
```diff
@@ -397,7 +397,20 @@ var data = [
 		"description" : "minimum of all of the numbers",
 		"example" : "(print (min 0.5 1 7 9 -5))"
 	},
-
+	{
+		"parameter" : "index_max *",
```
howsohazard: Should be "index_max [* x1] [number x2] ... [number xN]"
The description should be something similar to (fill in the missing info)
If x1 is a list or assoc, it will return the index of the maximum value ---- what happens if there is more than one? -----  Otherwise, it will return the index of the maximum value from among x1 through xN.

docs/language.js
```diff
@@ -397,7 +397,20 @@ var data = [
...
+		"output" : "*",
+		"concurrency" : true,
+		"description" : "the list index or assoc key mapped to the maximum value",
+		"example" : "(print (index_max 0.5 1 7 9 -5))\n(print (index_max (list 0.5 1 7 9 -5) ))\n(print (index_max (assoc 1 0.5 2 1 3 7)))"
+	},
+	{
+		"parameter" : "index_min *",
```
howsohazard: Update like index_max

src/Amalgam/interpreter/Interpreter.h
```diff
@@ -1309,6 +1311,88 @@ class Interpreter
...
+		if(value_found)
+			return  Parser::ParseFromKeyString(max_key->string, evaluableNodeManager);
+
+		return EvaluableNodeReference::Null();
+	}
+
+	template <typename Compare>
+	EvaluableNodeReference GetIndexMinMaxFromList(EvaluableNode *en, std::vector<EvaluableNode*> &orderedChildNodes, Compare compare, double compare_limit, bool immediate_result)
+	{
+
```
howsohazard, line 1336: remove extra newline

src/Amalgam/interpreter/InterpreterOpcodesMath.cpp
```diff
@@ -805,6 +805,36 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_MIN(EvaluableNode *en, boo
...
+{
+	auto &ocn = en->GetOrderedChildNodes();
+
+	if(ocn.size() == 0)
+		return EvaluableNodeReference::Null();
+
+	if(ocn[0] != nullptr && ocn[0]->GetType() == ENT_ASSOC)
+		return GetIndexMinMaxFromAssoc(ocn[0], std::greater(), -std::numeric_limits<double>::infinity(), immediate_result);
+	if(ocn[0] != nullptr && ocn[0]->GetType() == ENT_LIST)
+		return GetIndexMinMaxFromList(ocn[0], ocn[0]->GetOrderedChildNodes(), std::greater(), -std::numeric_limits<double>::infinity(), immediate_result);
```
howsohazard: On one hand, it'd be good for the function signatures to be consistent, on the other hand ocn has already been retrieved.  However, for the branch == ENT_LIST, can call GetOrderedChildNodesReference() since it's known to be a list.

src/Amalgam/interpreter/InterpreterOpcodesMath.cpp
```diff
@@ -805,6 +805,36 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_MIN(EvaluableNode *en, boo
...
+{
+	auto &ocn = en->GetOrderedChildNodes();
+
+	if(ocn.size() == 0)
+		return EvaluableNodeReference::Null();
+
+	if(ocn[0] != nullptr && ocn[0]->GetType() == ENT_ASSOC)
+		return GetIndexMinMaxFromAssoc(ocn[0], std::less(), std::numeric_limits<double>::infinity(), immediate_result);
+	if(ocn[0] != nullptr && ocn[0]->GetType() == ENT_LIST)
+		return GetIndexMinMaxFromList(ocn[0], ocn[0]->GetOrderedChildNodes(), std::less(), std::numeric_limits<double>::infinity(), immediate_result);
```
howsohazard: GetOrderedChildNodesReference here too

docs/language.js
```diff
@@ -397,7 +397,20 @@ var data = [
 		"description" : "minimum of all of the numbers",
 		"example" : "(print (min 0.5 1 7 9 -5))"
 	},
-
+	{
+		"parameter" : "index_max [number x1] [number x2] [number x3] ... [number xN] | (assoc ...) | (list ...)",
+		"output" : "*",
+		"concurrency" : true,
+		"description" : "the index of of the argument with the greatest value, or the key associated with the greatest value in an assoc, or the index of the greatest value in a list",
```
cademack: Should have looked into this PR earlier, sorry. I think I had put it in the ticket for this, but index_max/index_min need to return the *list* of indices that map to the maximum or minimum value.

This way the user is responsible for deciding the logic of which index to "get" in the case of ties within the structure. For example, the use-case that prompted me to ask for this would always desire the first index with the max value, but I can imagine other cases where I may want the last. I don't think these opcodes should only return one of the indices with the extreme value.

Many libraries only return the first, but in the case of (assoc)'s, the order is a bit hard to predict anyway.

naricc: Done.

docs/language.js
```diff
@@ -397,7 +397,20 @@ var data = [
 		"description" : "minimum of all of the numbers",
 		"example" : "(print (min 0.5 1 7 9 -5))"
 	},
-
+	{
+		"parameter" : "index_max [number x1] [number x2] [number x3] ... [number xN] | (assoc ...) | (list ...)",
+		"output" : "[*]",
+		"concurrency" : true,
+		"description" : "a list of the indices of the argument with the greatest value, or the keys associated with the greatest value in an assoc, or the indices of the greatest value in a list",
```
howsohazard: Since most of the opcodes are complete sentences and descriptions, should keep moving in that direction, improving readability.  Same with other opcode.

howsohazard: Also should improve clarity with regard to the three different forms of using it.

src/Amalgam/evaluablenode/EvaluableNode.cpp
```diff
@@ -1274,15 +1274,15 @@ void EvaluableNode::AppendOrderedChildNodes(const std::vector<EvaluableNode *> &
 	if(!IsOrderedArray())
 		return;
 
-	auto &ocn = GetOrderedChildNodesReference();
+	auto &ocn = GetOrderedChildNodesReference ();
```
howsohazard: No space between e and (.

src/Amalgam/evaluablenode/EvaluableNode.cpp
```diff
@@ -1274,15 +1274,15 @@ void EvaluableNode::AppendOrderedChildNodes(const std::vector<EvaluableNode *> &
...
 	ocn.insert(end(ocn), begin(ocn_to_append), end(ocn_to_append));
 
 	//if cycles, propagate upward
 	for(auto cn : ocn_to_append)
 	{
 		if(cn != nullptr && cn->GetNeedCycleCheck())
 		{
-			SetNeedCycleCheck(true);
+		SetNeedCycleCheck(true);
```
howsohazard: Indentation is off

src/Amalgam/evaluablenode/EvaluableNode.h
```diff
@@ -670,9 +670,9 @@ class EvaluableNode
 	__forceinline std::vector<EvaluableNode *> &GetOrderedChildNodes()
 	{
 		if(IsOrderedArray())
-			return GetOrderedChildNodesReference();
+			return GetOrderedChildNodesReference ();
```
howsohazard: No space between e and (

src/Amalgam/evaluablenode/EvaluableNode.h
```diff
@@ -670,9 +670,9 @@ class EvaluableNode
...
 
-		return emptyOrderedChildNodes;
+	return emptyOrderedChildNodes;
```
howsohazard: indentation is off

src/Amalgam/evaluablenode/EvaluableNode.h
```diff
@@ -886,8 +886,7 @@ class EvaluableNode
 	#if defined(MULTITHREAD_SUPPORT) || defined(MULTITHREAD_INTERFACE)
 		Concurrency::SingleLock lock(debugWatchMutex);
 	#endif
-		if(debugWatch.find(en) != end(debugWatch))
-		{
+		if(debugWatch.find(en) != end(debugWatch)) {
```
howsohazard: {'s go on their own line

src/Amalgam/interpreter/Interpreter.h
```diff
@@ -1308,6 +1310,130 @@ class Interpreter
 	//a stack (list) of the current nodes being executed
 	std::vector<EvaluableNode *> *opcodeStackNodes;
 
+	EvaluableNodeReference ConstructListWithZero();
+
+	EvaluableNodeReference IndexVectorToList(std::vector<size_t> indices, EvaluableNodeManager* evaluabNodeManager, bool immediate_result)
```
howsohazard: I'd recommend against this because there's often a lot of context in keeping track of freeing nodes and their flags, and it's not too much code to begin with.  Plus, this particular implementation creates 2 extra copies of unneeded lists.  And the EvaluableNodeReference constructor can be called with the allocated node, saving a line of code (as is frequently done elsewhere).

src/Amalgam/interpreter/Interpreter.h
```diff
@@ -1308,6 +1310,130 @@ class Interpreter
...
+				value_found = true;
+			}
+		}
+
+		if(value_found)
+		{
+			EvaluableNodeReference index_list;
+			index_list.SetReference(evaluableNodeManager->AllocNode(ENT_LIST));
+			auto &index_list_ocn = index_list->GetOrderedChildNodesReference();
+
```
howsohazard, line 1314: do a reserve here on the index_list_ocn to reduce memory allocations (improves CPU efficiency on large multicore machines)

src/Amalgam/interpreter/Interpreter.h
```diff
@@ -1308,6 +1310,130 @@ class Interpreter
...
+			for(StringInternPool::StringID max_key: max_keys)
+			{
```
howsohazard, line 1316: Normally wouldn't need {'s for one line for.  However, UpdatePropertiesBasedOnAttachedNode needs to be called in case the idem potent flag should be cleared due to a parsed index value.

src/Amalgam/interpreter/Interpreter.h
```diff
@@ -1308,6 +1310,130 @@ class Interpreter
...
+			return EvaluableNodeReference::Null();
+
+		bool value_found = false;
+		double result_value = compare_limit;
+
+	#ifdef MULTITHREAD_SUPPORT
+		std::vector<EvaluableNodeReference> interpreted_nodes;
+		if(InterpretEvaluableNodesConcurrently(en, orderedChildNodes, interpreted_nodes, true))
+		{
+			std::vector<size_t> max_indices;
```
howsohazard: can declare this up top for both paths

src/Amalgam/interpreter/InterpreterOpcodesMath.cpp
```diff
@@ -1,11 +1,11 @@
 //project headers:
 #include "Interpreter.h"
-
+#include "EvaluableNode.h"
```
howsohazard: EvaluableNode.h goes before Interpreter.h due to alphabetical order.

src/Amalgam/interpreter/InterpreterOpcodesMath.cpp
```diff
@@ -805,6 +805,60 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_MIN(EvaluableNode *en, boo
...
+	if(ocn.size() == 1) 
+	{
+		EvaluableNodeReference ocn_zero = InterpretNode(ocn[0]);
+
+		if(ocn_zero != nullptr && ocn_zero->GetType() == ENT_ASSOC)
+			return GetIndexMinMaxFromAssoc(ocn_zero, std::greater(), -std::numeric_limits<double>::infinity(), immediate_result);
+		else if(ocn_zero != nullptr && ocn_zero->GetType() == ENT_LIST)
+			return GetIndexMinMaxFromList(ocn_zero, ocn_zero->GetOrderedChildNodesReference(), std::greater(), -std::numeric_limits<double>::infinity(), immediate_result);
+		else
+			return ConstructListWithZero();
```
howsohazard: Seems like an odd/unusual outcome to return a list of zero.  I think the following logic would be more consistent: 
if assoc, do assoc
else if immediate, do list on all the nodes
else must be list, do list on the list

howsohazard: Then don't need ConstructListWithZero either.

src/Amalgam/interpreter/InterpreterOpcodesMath.cpp
```diff
@@ -805,6 +805,60 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_MIN(EvaluableNode *en, boo
...
+}
+
+EvaluableNodeReference Interpreter::InterpretNode_ENT_INDEX_MIN(EvaluableNode *en, bool immediate_result)
+{
+	auto &ocn = en->GetOrderedChildNodes();
+
+	if(ocn.size() == 0)
+		return EvaluableNodeReference::Null();
+
+	if(ocn.size() == 1)
```
howsohazard: Same logic as index_max comment above

src/Amalgam/interpreter/InterpreterOpcodesMath.cpp
naricc: So the reason I did it this way is because, in order to know the type of the first child node, I need to evaluate it (i.e. call InterpretNode on it). So if there is exactly one child node, and its not a list or assoc, the max index must be 0 (because its the only index).

src/Amalgam/interpreter/Interpreter.h
naricc: You mean don't make this a separate method?

src/Amalgam/interpreter/Interpreter.h
```diff
@@ -1308,7 +1310,134 @@ class Interpreter
 	//a stack (list) of the current nodes being executed
 	std::vector<EvaluableNode *> *opcodeStackNodes;
 
-public:
+	EvaluableNodeReference IndexVectorToList(std::vector<size_t>& indices, EvaluableNodeManager *evaluableNodeManager, bool immediate_result)
```
howsohazard: space before &

src/Amalgam/interpreter/Interpreter.h
```diff
@@ -1308,7 +1310,134 @@ class Interpreter
...
+	{
+		EvaluableNodeReference index_list(evaluableNodeManager->AllocNode(ENT_LIST), false);
+		std::vector<EvaluableNode *> &index_list_ocn = index_list->GetOrderedChildNodesReference();
```
howsohazard: can use auto & here

src/Amalgam/interpreter/Interpreter.h
```diff
@@ -1308,7 +1310,134 @@ class Interpreter
...
+			index_list_ocn.push_back(AllocReturn(static_cast<double>(index), immediate_result));
+		}
+
+		return index_list;
+	}
+
+	template <typename Compare>
+	EvaluableNodeReference GetIndexMinMaxFromAssoc(EvaluableNodeReference interpreted_assoc, Compare compare, double compare_limit, bool immediate_result)
+	{
+		EvaluableNode::AssocType mapped_child_nodes = interpreted_assoc->GetMappedChildNodesReference();
```
howsohazard: This makes an unneeded copy -- should do auto &

src/Amalgam/interpreter/Interpreter.h
```diff
@@ -1308,7 +1310,134 @@ class Interpreter
...
+				candidate_value = cur_value;
+				max_keys.push_back(cur_key);
+				value_found = true;
+			}
+		}
+
+		if(value_found)
+		{
+			EvaluableNodeReference index_list;
+			index_list.SetReference(evaluableNodeManager->AllocNode(ENT_LIST));
```
howsohazard: Combine these two lines, just use EvaluableNodeReference constructor

src/Amalgam/interpreter/Interpreter.h
```diff
@@ -1308,7 +1310,134 @@ class Interpreter
...
+	}
+
+	template <typename Compare>
+	EvaluableNodeReference GetIndexMinMaxFromList(EvaluableNode *en, Compare compare, double compare_limit, bool immediate_result)
+	{
+		bool value_found = false;
+		double result_value = compare_limit;
+		std::vector<size_t> max_indices;
+
+		std::vector<EvaluableNode *> orderedChildNodes = en->GetOrderedChildNodesReference();
```
howsohazard: Use auto & -- an unintended copy is being made here

src/Amalgam/interpreter/InterpreterOpcodesMath.cpp
```diff
@@ -805,6 +805,40 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_MIN(EvaluableNode *en, boo
...
 }
 
+EvaluableNodeReference Interpreter::InterpretNode_ENT_INDEX_MAX(EvaluableNode *en, bool immediate_result)
+{
+	auto &ocn = en->GetOrderedChildNodes();
+
+	if(ocn.size() == 0)
+		return EvaluableNodeReference::Null();
+
+	EvaluableNodeReference ocn_zero = InterpretNode(ocn[0], immediate_result);
```
howsohazard: In this implementation, the first node will be interpreted twice, which can be problematic if there are side effects.  Will need to move the concurrent evaluation here instead of leaving it in the list method.

src/Amalgam/interpreter/Interpreter.h
```diff
@@ -1308,7 +1310,134 @@ class Interpreter
...
+	{
+		EvaluableNode::AssocType mapped_child_nodes = interpreted_assoc->GetMappedChildNodesReference();
+		double candidate_value = compare_limit;
+		bool value_found = false;
+
+		std::vector<StringInternPool::StringID> max_keys;
+
+		for(auto [cur_key, cur_child] : mapped_child_nodes)
+		{
+			double cur_value = InterpretNodeIntoNumberValue(cur_child);
```
howsohazard: No need to interpret again, since this node has already been interpreted.

src/Amalgam/interpreter/InterpreterOpcodesMath.cpp
```diff
@@ -805,6 +805,40 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_MIN(EvaluableNode *en, boo
...
+
+	if(ocn_zero != nullptr && ocn_zero->GetType() == ENT_ASSOC && ocn.size() == 1)
+		return GetIndexMinMaxFromAssoc(ocn_zero, std::greater(), -std::numeric_limits<double>::infinity(), immediate_result);
+	else if(ocn_zero != nullptr && ocn_zero->GetType() == ENT_LIST && ocn.size() == 1)
+		return GetIndexMinMaxFromList(ocn_zero, std::greater(), -std::numeric_limits<double>::infinity(), immediate_result);
+	else
+		return GetIndexMinMaxFromList(en, std::greater(), -std::numeric_limits<double>::infinity(), immediate_result);
+}
```
howsohazard, line 827: Should free ocn_zero if possible once it is no longer needed (certainly before end of this method)

src/Amalgam/interpreter/InterpreterOpcodesMath.cpp
```diff
@@ -805,6 +805,40 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_MIN(EvaluableNode *en, boo
...
+
+EvaluableNodeReference Interpreter::InterpretNode_ENT_INDEX_MIN(EvaluableNode *en, bool immediate_result)
+{
```
howsohazard, line 830: See comments for index_max

src/Amalgam/interpreter/Interpreter.h
naricc: I made the indices pass by reference.

src/Amalgam/interpreter/Interpreter.h
```diff
@@ -1308,7 +1310,186 @@ class Interpreter
 	//a stack (list) of the current nodes being executed
 	std::vector<EvaluableNode *> *opcodeStackNodes;
 
-public:
+	EvaluableNodeReference IndexVectorToList(std::vector<size_t> &indices, EvaluableNodeManager *evaluableNodeManager, bool immediate_result)
```
howsohazard: No need for this method, it already exists: CreateListOfNumbersFromIteratorAndFunction

src/Amalgam/interpreter/Interpreter.h
```diff
@@ -1308,7 +1310,186 @@ class Interpreter
...
+			index_list_ocn.push_back(AllocReturn(static_cast<double>(index), immediate_result));
+		}
+
+		return index_list;
+	}
+
+	template <typename Compare>
+	EvaluableNodeReference GetIndexMinMaxFromAssoc(EvaluableNodeReference interpreted_assoc, Compare compare, double compare_limit, bool immediate_result)
+	{
+		auto& mapped_child_nodes = interpreted_assoc->GetMappedChildNodesReference();
```
howsohazard: Space before the &, not after.

src/Amalgam/interpreter/Interpreter.h
howsohazard: Still an issue -- in the opcode methods (Interpreter::InterpretNode_ENT_INDEX_MAX), there's an interpret for immediate use.  So these all are already interpreted.  Same applies to other code paths below (including for list and the concurrent implementation).

src/Amalgam/Opcodes.h
```diff
@@ -313,7 +315,7 @@ constexpr OrderedChildNodeType GetInstructionOrderedChildNodeType(EvaluableNodeT
 	case ENT_PARALLEL:
 	case ENT_ADD:
 	case ENT_MULTIPLY:
-	case ENT_MAX:					case ENT_MIN:
+	case ENT_MAX:					case ENT_MIN:				case ENT_INDEX_MAX:	case ENT_INDEX_MIN:
 	case ENT_AND:					case ENT_OR:				case ENT_XOR:
 	case ENT_EQUAL:					case ENT_NEQUAL:
```
howsohazard, line 320: Need to update the branch to latest main and include ENT_INDEX_MAX/MIN in the new GetOpcodeNewValueReturnType method.

naricc: Rebased.

src/Amalgam/interpreter/Interpreter.h
```diff
@@ -1274,7 +1276,123 @@ class Interpreter
...
+	}
+
+	template <typename Compare>
+	EvaluableNodeReference GetIndexMinMaxFromRemainingArgList(EvaluableNode *en, Compare compare, double compare_limit, bool immediate_result)
+	{
+		double result_value = compare_limit;
+		std::vector<size_t> max_indices;
+
+		// If this method is called, we have already Intepreted the first arugment
+		max_indices.push_back(0);
```
howsohazard: This behavior is not consistent with the other paths in the case that there is a null value which is interpreted as a NaN.  Suppose you have only null (NaN) values, in that case there is no maximum value and so it will return null in the other paths, but in this one it will return the first index.

src/Amalgam/interpreter/InterpreterOpcodesMath.cpp
```diff
@@ -804,6 +804,50 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_MIN(EvaluableNode *en, boo
...
+	EvaluableNodeReference ocn_zero = InterpretNodeForImmediateUse(ocn[0]);
+	auto node_stack = CreateOpcodeStackStateSaver(ocn_zero);
+
+	EvaluableNodeReference result;
+	if(ocn_zero != nullptr && ocn_zero->GetType() == ENT_ASSOC && ocn.size() == 1)
+		result = GetIndexMinMaxFromAssoc(ocn_zero, std::greater(), -std::numeric_limits<double>::infinity(), immediate_result);
+	else if(ocn_zero != nullptr && ocn_zero->GetType() == ENT_LIST && ocn.size() == 1)
+		result = GetIndexMinMaxFromList(ocn_zero, std::greater(), -std::numeric_limits<double>::infinity(), immediate_result);
+	else
+		return GetIndexMinMaxFromRemainingArgList(en, std::greater(), ConvertNodeIntoNumberValueAndFreeIfPossible(ocn_zero), immediate_result);
```
howsohazard: Calling ConvertNodeIntoNumberValueAndFreeIfPossible here and then FreeNodeTreeIfPossible below can result in a double free, which can cause problems.  Additionally, the other nodes have not been interpreted in this path, so within GetIndexMinMaxFromRemainingArgList, the numeric values should be obtained via InterpretNodeIntoNumberValue.

src/Amalgam/interpreter/Interpreter.h
```diff
@@ -1363,33 +1366,45 @@ class Interpreter
...
+		bool value_found = false;
 
 		auto &orderedChildNodes = en->GetOrderedChildNodesReference();
 
 		if(orderedChildNodes.size() == 0)
 			return EvaluableNodeReference::Null();
 
+		// First node has already been interpreted and thus needs different handling
+		double candidate_zero = EvaluableNode::ToNumber(orderedChildNodes[0]);
+		if(!std::isnan(candidate_zero))
```
howsohazard: Use FastIsNaN, as is used all over.  std::isnan is remarkably slow due to weird historical reasons.

src/Amalgam/amlg_code/full_test.amlg
```diff
@@ -399,6 +399,7 @@
  (print (index_max .5 -12 3 5 7))
  (print (index_max (list 1 1 3 2 1 3)))
  (print (index_max (null) 34  -66))
+ (print (index_max "NaN" "NaN" "NaN") "\n")
```
howsohazard: This is testing string comparisons, not NaNs.  For reading clarity, I'd recommend either using strings like "a" "b" "c" or use (null)'s that resolve to NaNs

src/Amalgam/interpreter/Interpreter.h
```diff
@@ -1363,33 +1366,45 @@ class Interpreter
...
+
 		for(size_t i = 1; i < orderedChildNodes.size(); i++)
 		{
-			double cur_value = EvaluableNode::ToNumber(orderedChildNodes[i]);
+			double cur_value = InterpretNodeIntoNumberValue(orderedChildNodes[i]);
 
 			if(cur_value == result_value)
 			{
 				max_indices.push_back(i);
+				value_found = true;
```
howsohazard, line 1398: Is this set to true needed?  Seems like it shouldn't be, and the fact that it is here makes the reader question what's going on.

src/Amalgam/interpreter/InterpreterOpcodesMath.cpp
```diff
@@ -1,11 +1,11 @@
 //project headers:
-#include "Interpreter.h"
-
+#include "EvaluableNode.h"
```
howsohazard: EvaluableNode.h should come after EntityQueryBuilder.h alphabetically

src/Amalgam/interpreter/Interpreter.h
naricc: It is needed for an edge case where orderedChildNodes  is a list of all `std::numeric_limits<double>::infinity()` or (-std::numeric_limits<double>::infinity() in the case of min). If that happens we never fall into the other case, and this function would return `return EvaluableNodeReference::Null();`.

I can put a comment to clarify this.

src/Amalgam/Opcodes.h
howsohazard: You'll need to include ENT_INDEX_MAX/MIN in the GetOpcodeNewValueReturnType .

src/Amalgam/interpreter/Interpreter.h
naricc: Added the comment.

howsohazard (approved):

PR #318, opened 2024-12-04
22367: Fixes deadlock issue with entity writes

howsohazard:


jdbeel (commented):
Should this be MAJOR? I'm not 100% up on the versioning practices in this repo but `CollectGarbage` being renamed is a non-backwards-compatible change.

EDIT: I see that `CollectGarbage` still exists, but I see that some arguments were removed so perhaps it's just MINOR?

src/Amalgam/interpreter/InterpreterOpcodesEntityControl.cpp
```diff
@@ -197,11 +197,7 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_ASSIGN_ENTITY_ROOTS_and_AC
 			}
 		}
 
-	#ifdef MULTITHREAD_SUPPORT
-		target_entity->CollectGarbage(&memoryModificationLock);
-	#else
-		target_entity->CollectGarbage();
-	#endif
+		target_entity->CollectGarbageWithEntityWriteReference();
```
naricc, line 200: So it seems like the difference is basically we used to hold a write lock during this garbage collection; with this change that is no longer the case. But why is it ok to not hold that lock here?

howsohazard: Different things are locked.  In general, there's a global lock for garbage collection so only one collect can happen at a time -- that's a static attribute of EvaluableNodeManager.  However, there's already a write lock on the entity itself. The deadlock issue is that the write lock on target_entity was preventing any read locks in other threads, which were preventing them from ever going into garbage collection.  But that was masking another issue in that they wouldn't go into garbage collection until the entities that they are in were in need.  So even just fixing the immediate issue would hold this thread idle and waste CPU resources.

Instead, this new approach uses a different heuristic which additionally looks to see if anything is executing on the target_entity.  If so, it'll just let those threads handle the garbage collection.  If not, then the write lock will prevent any other thread from calling this entity, so it doesn't need to do any additional lock.

naricc: Oh, I see, target_entiy is a WriteReference which itself has a lock. And since GarbageCollection is specific to one EvaluableNode, that is sufficient. Do I have that right?

naricc (approved):

src/Amalgam/interpreter/InterpreterOpcodesEntityControl.cpp
howsohazard: To one Entity, but yes!

PR #324, opened 2024-12-11
22250: Adds EvalOnEntity to Amalgam API, MINOR

dmaze:
`EvalOnEntity(handle, "(arbitrary_code)")` runs that code in the context of some entity.

### DIFF
--- a/src/Amalgam/Amalgam.h
+++ b/src/Amalgam/Amalgam.h
@@ -60,6 +60,8 @@ extern "C"
 	AMALGAM_EXPORT wchar_t *ExecuteEntityJsonPtrWide(char *handle, char *label, char *json);
 	AMALGAM_EXPORT char *ExecuteEntityJsonPtr(char *handle, char *label, char *json);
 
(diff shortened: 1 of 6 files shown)


src/Amalgam/Amalgam.h
```diff
@@ -60,6 +60,8 @@ extern "C"
 	AMALGAM_EXPORT wchar_t *ExecuteEntityJsonPtrWide(char *handle, char *label, char *json);
 	AMALGAM_EXPORT char *ExecuteEntityJsonPtr(char *handle, char *label, char *json);
 
+	AMALGAM_EXPORT char *EvalEntity(char *handle, char *amlg);
```
howsohazard: I can't help but to think there should be another word here.  ExecuteEntity means it will run something /of/ the entity.  I almost feel like this could be ExecuteOnEntity, or EvalOnEntity, or something.  Thoughts?

dmaze: I like "eval" here (because you're injecting code directly), changing it to "...on entity" isn't a big deal, done.

howsohazard (approved):

PR #326, opened 2024-12-16
22408: Add ExecuteEntityJsonPtrLogged API call

dmaze:
Returns both the JSON data, as in ExecuteEntityJsonPtr, and also a transaction-log entry that can be sent back via the EvalOnEntity API.

### DIFF
--- a/test/lib_smoke_test/main.cpp
+++ b/test/lib_smoke_test/main.cpp
@@ -6,37 +6,293 @@
 #include "Amalgam.h"
 
 //system headers:
+#include <functional>
 #include <iostream>
 #include <string>
 
-int main(int argc, char* argv[])
+// A wrapper around a C string that DeleteString() on exit.
+class ApiString
 {
-	// Print version:
-	std::cout << std::string(GetVersionString()) << std::endl;
+	char *p;
 
+public:
+	ApiString(char *p) : p(p) {}
+	ApiString(const ApiString &a) = delete;
+	ApiString(ApiString &&a) : p(a.p)
+	{
+		a.p = nullptr;
+	}
+
+	~ApiString()
+	{
+		if(p != nullptr)
+			DeleteString(p);
+	}
+
+	operator std::string() const
+	{
+		std::string s(p);
+		return s;
+	}
+};
+
+// A wrapper around an entity that DestroyEntity() on exit.
+class LoadedEntity
+{
+	std::string h;
+
+public:
+	LoadedEntity(const std::string &handle) : h(handle) {}
+
+	~LoadedEntity()
+	{
+		DestroyEntity(h.data());
+	}
+
+	const std::string &handle() const
+	{
+		return h;
+	}
+};
+
+class TestResult
+{
+	std::string test;
+	bool successful;
+
+public:
+	TestResult(const std::string &test) : test(test), successful(true) {}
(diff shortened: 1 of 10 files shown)


naricc (approved):

test/catch2_test/CMakeLists.txt
```diff
@@ -0,0 +1,59 @@
...
+    # error LNK2001: unresolved external symbol main
+    message(STATUS "Will not run Catch2 tests on Windows")
+else()
+
+    # Get the Catch2 unit test framework from its GitHub repository
+    Include(FetchContent)
+    FetchContent_Declare(
+        Catch2
+        GIT_REPOSITORY https://github.com/catchorg/Catch2.git
+        GIT_TAG        v3.7.1
```
howsohazard: Instead of making the build process more complex, we could include the amalgamated cpp & header from their releases: https://github.com/catchorg/Catch2/releases

dmaze: The Catch2 documentation seems to [recommend avoiding the amalgated files as a best practice](https://github.com/catchorg/Catch2/blob/devel/docs/migrate-v2-to-v3.md#how-to-migrate-projects-from-v2-to-v3).  The fragment you're highlighting here is [directly out of the Catch2 CMake documentation](https://github.com/catchorg/Catch2/blob/devel/docs/cmake-integration.md#cmake-targets).  Pretty much everything outside this immediate block would stay the same whether we used `FetchContent` or bundled a copy of the source ourselves.

I think using `FetchContent` is going to be easier to maintain (what are the versions of the `src/3rd_party` libraries?) and avoids bundling somebody else's code into our own repository.

dmaze: We had some discussion around this, and concluded that none of the paths to wiring Catch2 into the system were good ones.  (We'd prefer a self-contained repository; cloning Catch2 itself is a lot of code; the "amalgamated" release is two large files and appears to not be the preferred path.). I replaced this setup with a bespoke test framework for the top-level API only.

test/lib_smoke_test/main.cpp
```diff
@@ -6,37 +6,284 @@
 #include "Amalgam.h"
 
 //system headers:
+#include <functional>
 #include <iostream>
 #include <string>
 
-int main(int argc, char* argv[])
+// A wrapper around a C string that DeleteString() on exit.
+class ApiString {
```
howsohazard: style: curlies go on next line

test/lib_smoke_test/main.cpp
```diff
@@ -6,37 +6,284 @@
...
+	char *p;
+
+	public:
```
howsohazard: public/protected/etc. are back a tab

test/lib_smoke_test/main.cpp
```diff
@@ -6,37 +6,284 @@
...
+	ApiString(char *p) : p(p) {}
+	ApiString(const ApiString &a) = delete;
+	ApiString(ApiString &&a) : p(a.p) {
+		a.p = nullptr;
+	}
+
+	~ApiString()
+	{
+		if(p != nullptr)
+		{
```
howsohazard: don't need curlies when there's a singular line after an if

PR #327, opened 2024-12-16
22286: Amlg lang improvements: removes query_weighted_sample, weighted_rand, updates rand, query_sample, sort, reduce, MAJOR

howsohazard:
Removes query_weighted_sample and weighted_rand, updates query_sample and rand to handle functionality.
Removes access to (target) from reduce and sort opcodes for performance and consistency reasons.
Improves documentation and clarifies some opcode use.


src/Amalgam/entity/EntityQueries.cpp
```diff
@@ -352,75 +351,53 @@ EvaluableNodeReference EntityQueryCondition::GetMatchingEntities(Entity *contain
...
-			matching_entities.clear();
-			return EvaluableNodeReference::Null();
-		}
-
-		//retrieve weights
-		std::vector<double> entity_weights;
-		entity_weights.reserve(num_to_sample);
+			//retrieve weights
+			std::vector<double> entity_weights;
+			entity_weights.reserve(num_to_sample);
```
dmaze, line 368: (`std::vector<T>` has a constructor that takes the desired size, so this could be `std::vector<double> entity_weights(num_to_sample)`.)

src/Amalgam/entity/EntityQueries.cpp
```diff
@@ -352,75 +351,53 @@ EvaluableNodeReference EntityQueryCondition::GetMatchingEntities(Entity *contain
...
 
-		//retrieve and accumulate weights
-		for(size_t i = 0; i < matching_entities.size(); i++)
-		{
-			double value;
-			Entity *e = matching_entities[i];
-			if(e != nullptr && e->GetValueAtLabelAsNumber(weight_label_id, value))
+			//retrieve and accumulate weights
+			for(size_t i = 0; i < matching_entities.size(); i++)
```
dmaze: (Should this use the C++ iterator pattern;

```c++
for(auto iter = matching_entities.begin(); iter != matching_entities.end(); iter++)
{
  Entity *e = *iter;
  ...
}
```

There are also functional-style `std::for_each()` and `std::transform()`, but the C++ lambda syntax is a little tricky, and I have no idea what their performance implications are.)

dmaze (approved):

src/Amalgam/entity/EntityQueries.cpp
howsohazard: Very ancient code here just moved around, updated to use range based for

src/Amalgam/entity/EntityQueries.cpp
howsohazard: Want to use reserve here as the constructor will automatically initialize (which isn't desired or needed since the push_back will do so below)

PR #331, opened 2024-12-29
22414: Improves memory reuse of apply opcode, improves profiling report, and improves AFMI's ability to catch potential memory corruption

howsohazard:


docs/language.js
```diff
@@ -144,6 +144,113 @@ var data = [
...
+		"example" : "(print (set (list 1 2 3 4) 2 7))\n(print (set\n  (list (assoc \"a\" 1))\n    (list 2) 1\n    (list 1) (get (target) 0)))"
+	},
+
+	{
+		"parameter" : "replace * data [number index1|string index1|list walk_path1] [* function1] [number index2|string index2|list walk_path2] [* function2] ...",
```
howsoRes, line 164: aren't all these opcodes already in language.js ? why is it the diff showing these all being added as new definitions when they already exist?

howsohazard: Because they were moved from below.

howsoRes (approved):

PR #332, opened 2024-12-31
22045: Implements undefined variable warnings and performance constraint warnings, MAJOR

naricc:
Adds undefined variable warnings and perf constraint warnings  in amalgam. 

The will return a tuple of the form [result, warnings, perf_constraint_violation]
call_sandboxed
call_entity
call_entity_get_changes
call_container


src/Amalgam/interpreter/InterpreterOpcodesBase.cpp
```diff
@@ -577,9 +579,64 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_CALL_SANDBOXED(EvaluableNo
...
+																							ConstraintViolationToString);													
+
+	EvaluableNodeReference result_tuple(evaluableNodeManager->AllocNode(ENT_LIST), true);
+
+	auto &result_tuple_ocn = result_tuple->GetOrderedChildNodesReference();
+
+	result_tuple_ocn.reserve(3);
+	result_tuple_ocn.push_back(result);
+	result_tuple_ocn.push_back(warning_assoc);
+	result_tuple_ocn.push_back(constraint_violation_list);
```
howsohazard: Need to update result_tuple node flags for all of the new child nodes.

src/Amalgam/interpreter/Interpreter.h
```diff
@@ -129,6 +132,14 @@ class PerformanceConstraints
...
 	//If 0, then unlimited
 	size_t maxEntityIdLength;
+
+	//If true, collectWarnings
+	bool collectWarnings;
+
+	//maps warnings with their occurance count, if collectWarnings is true 
+	FastHashMap<std::string, size_t> warnings;
+
+	FastHashSet<ViolationType> constraintViolations;
```
howsohazard: Since the first performance constraint that fails will stop execution, it doesn't need to be a set, just a single value

src/Amalgam/interpreter/Interpreter.cpp
```diff
@@ -309,6 +309,7 @@ Interpreter::Interpreter(EvaluableNodeManager *enm, RandomStream rand_stream,
 	constructionStackNodes = nullptr;
 
 	evaluableNodeManager = enm;
+	warnOnUndefined = warn_on_undefined;
```
howsohazard: just lean on the performance constraints object having a flag for collecting warnings, don't need one in Interpreter

src/Amalgam/entity/Entity.h
```diff
@@ -260,14 +260,14 @@ class Entity
 	// If direct_get is true, then it will return values with all labels
 	// If on_self is true, then it will be allowed to access private variables
 	// If batch_call is true, then it assumes it will be called in a batch of updates and will not perform any cleanup or synchronization
-	EvaluableNodeReference GetValueAtLabel(StringInternPool::StringID label_sid, EvaluableNodeManager *destination_temp_enm, bool direct_get,
+	std::pair<EvaluableNodeReference, bool>  GetValueAtLabel(StringInternPool::StringID label_sid, EvaluableNodeManager *destination_temp_enm, bool direct_get,
```
howsohazard: Need to implement the same return type pattern for GetValueAtLabelAs* methods.  Make sure to update your branch since there have been changes to the immediate one.

src/Amalgam/interpreter/Interpreter.h
naricc: So it seemed like you could have multiple threads, and potentially N of them could run into a performance constraint before any of the others manage to stop execution. Is that actually impossible? I agree its kind of a weird edge case; I could just return whichever one happens to hit first.

howsohazard: I think just return whichever hit first.

src/Amalgam/AssetManager.h
```diff
@@ -566,6 +566,10 @@ class AssetManager
 	//if true, will enable debugging the sources of loading nodes
 	bool debugSources;
 
+	//if true, the interpreter will print warnings on undefined variables to standard error.
+	//If debugSources is true these warnings will (additionally) contain debug source information.
+	bool warnOnUndefined;
```
howsohazard, line 574: Need to default this to false somewhere.

src/Amalgam/entity/EntityQueries.cpp
```diff
@@ -371,7 +371,18 @@ EvaluableNodeReference EntityQueryCondition::GetMatchingEntities(Entity *contain
...
+				double found;
+
+				if(e != nullptr)
+				{
+					auto value_found = e->GetValueAtLabelAsNumber(singleLabel);
+					value = value_found.first;
+					found = value_found.second;
+				}
+				else
+					found = false;
```
howsohazard, line 379: Style: braces around this because there were braces around the above if

src/Amalgam/AssetManager.h
```diff
@@ -566,6 +566,10 @@ class AssetManager
 	//if true, will enable debugging the sources of loading nodes
 	bool debugSources;
 
+	//if true, the interpreter will print warnings on undefined variables to standard error.
+	//If debugSources is true these warnings will (additionally) contain debug source information.
+	bool warnOnUndefined = false;
```
howsohazard: Since this isn't a constant, would be better to set up in the constructor with the rest of the initialization, as is done throughout the rest of the code.

docs/language.js
```diff
@@ -1711,31 +1711,31 @@ var data = [
...
...ory.   If max_node_allocations is 0 or infinite, then there is no limit to the number of nodes to be allotted to the entity as long as the machine has sufficient memory, but only if the containing entity (the current entity) has unlimited memory access.  If max_opcode_execution_depth is 0 or infinite and the caller also has no limit, then there is no limit to the depth that opcodes can execute, otherwise max_opcode_execution_depth limits how deep nested opcodes will be called.  The parameters max_contained_entities, max_contained_entity_depth, and max_entity_id_length constrain what they describe, and are primarily useful when ensuring that an entity and all its contained entities can be stored out to the filesystem.  The execution performed will use a random number stream created from the entity's random number stream. If the optional return_warnings parameter is true (the default), the result will be a tuple of the form [value, warnings, performance_constraint_violation], where warnings is a list of all undefined variable warnings, and perf_constraint violation is a string denoting the performance violation encounter (empty string if none). If false, only the value is returned.",
```
howsohazard: If this defaults to true, this means we'll need a large number of code changes in many places where we don't care about warnings.  Default should be conditional based on whether any of the constraints are set -- if no constraints are set, then it defaults to false and just returns the value, if any constraints are set, then it defaults to true and returns the tuple.  That way you can force it to etiher true or false.  Should make this consistent across all of the methods.  I recall this is what we discussed and agreed upon.

src/Amalgam/AssetManager.h
howsohazard: Still need to initialize it in the constructor.  The main method is not called if Amalgam is loaded as a .dll/.so/.dylib.

src/Amalgam/interpreter/Interpreter.h
```diff
@@ -16,6 +16,7 @@
 #include <string>
 #include <type_traits>
 #include <vector>
+#include <unordered_set>
```
howsohazard: Unneeded include.

src/Amalgam/interpreter/Interpreter.h
```diff
@@ -130,6 +133,32 @@ class PerformanceConstraints
 	//constrains the maximum length of an entity id (primarily to make sure it doesn't cause problems for file systems)
 	//If 0, then unlimited
 	size_t maxEntityIdLength;
+
+	//If true, collectWarnings
+	bool collectWarnings;
+
+	void addWarning(std::string warning)
```
howsohazard: Following style, methods occur before variables in class ordering.  Also methods shoult start with a captial letter like types.

src/Amalgam/interpreter/Interpreter.h
```diff
@@ -130,6 +133,32 @@ class PerformanceConstraints
...
+#endif
+		warnings[warning]++;
+	}
+
+	FastHashMap<std::string, size_t> &GetWarnings()
+	{
+		return warnings;
+	}
+
+	ViolationType constraintViolation = ViolationType::NoViolation;
```
howsohazard: This is inconsistent with how the rest of the variables are initialized.  Either should move this there or move all initialization to a constructor.

src/Amalgam/AmalgamMain.cpp
```diff
@@ -165,6 +166,8 @@ PLATFORM_MAIN_CONSOLE
 		}
 		else if(args[i] == "--debug-sources")
 			debug_sources = true;
+		else if(args[i] == "--no-warn-on-undefined")
```
howsohazard, line 172: Need to add this to the help output above.

src/Amalgam/interpreter/InterpreterOpcodesDataTypes.cpp
```diff
@@ -194,10 +195,29 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_SYMBOL(EvaluableNode *en,
 			return value;
 	}
 
-	//if didn't find it in the stack, try it in the labels
+	// if didn't find it in the stack, try it in the labels
 	EntityReadReference cur_entity_ref(curEntity);
 	if(cur_entity_ref != nullptr)
-		return cur_entity_ref->GetValueAtLabel(sid, nullptr, true, true);
+	{
+		std::pair<EvaluableNodeReference, bool> value_tuple = cur_entity_ref->GetValueAtLabel(sid, nullptr, true, true);
```
howsohazard: why not use auto [] here?

src/Amalgam/interpreter/InterpreterOpcodesDataTypes.cpp
```diff
@@ -194,10 +195,29 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_SYMBOL(EvaluableNode *en,
...
+			{
+				if(performanceConstraints != nullptr)
+					performanceConstraints->addWarning(sid->string + " " + en->GetCommentsString());
+			}
+		}
+
+		return cur_entity_ref->GetValueAtLabel(sid, nullptr, true, true).first;
+	}
+
+	if (performanceConstraints != nullptr && performanceConstraints->collectWarnings)
```
howsohazard: not conformant to style with space after if

src/Amalgam/AssetManager.h
naricc: I've moved the initialization to the constructor.

src/Amalgam/entity/EntityQueries.cpp
naricc: Done.

src/Amalgam/AssetManager.h
naricc: Moved initialization to the constructor.

docs/language.js
naricc: I've changed the behavior so that, if not constraints are set, it returns just the value. If constraints are set, return the tuple, unless return_warnings is false.

src/Amalgam/interpreter/Interpreter.h
naricc: Removed.

src/Amalgam/interpreter/Interpreter.h
naricc: Reordered.

src/Amalgam/interpreter/Interpreter.h
naricc: I've moved initialization to PopulatePerformanceConstraintsFromParams, which seems to be where the others are initialized.

src/Amalgam/interpreter/InterpreterOpcodesDataTypes.cpp
naricc: Fixed.

src/Amalgam/AmalgamMain.cpp
naricc: Added.

src/Amalgam/interpreter/InterpreterOpcodesDataTypes.cpp
naricc: Changed to use the auto pattern.

docs/language.js
```diff
@@ -30,7 +30,7 @@ var data = [
...
+		"description" : "String is parsed into code, and the result is returned.  If transactional is false, the default, it will attempt to parse the whole string and will return the closest code possible if there are any parse issues.  If transactional is true, it will parse the string transactionally, meaning that any node that has a parse error or is incomplete will be omitted along with all child nodes except for the top node.  If any performance constraints are given, the result will be a tuple of the form [value, warnings, performance_constraint_violation], where warnings is a list of all undefined variable warnings, and perf_constraint violation is a string denoting the constraint exceeded (empty string if none), unless return_warnings is false, in which case just the value will be returned.",
```
howsohazard: If any performance constraints are given or return_warnings is true,...

howsohazard: And update others below too.

src/Amalgam/entity/EntityQueries.cpp
```diff
@@ -371,7 +371,20 @@ EvaluableNodeReference EntityQueryCondition::GetMatchingEntities(Entity *contain
 			for(Entity *e : matching_entities)
 			{
 				double value;
-				if(e != nullptr && e->GetValueAtLabelAsNumber(singleLabel, value))
+				double found;
+
+				if(e != nullptr)
+				{
+					auto value_found = e->GetValueAtLabelAsNumber(singleLabel);
```
howsohazard: You could use std::tie to make this code more compact:
std::tie(value, found) = ...

src/Amalgam/entity/EntityQueries.cpp
```diff
@@ -600,17 +622,24 @@ EvaluableNodeReference EntityQueryCondition::GetMatchingEntities(Entity *contain
...
 			};
 
 			auto get_weight = [matching_entities, this]
 			(size_t i, double &weight_value)
 			{
-				return matching_entities[i]->GetValueAtLabelAsNumber(weightLabel, weight_value);
+				auto [ret_val, found] = matching_entities[i]->GetValueAtLabelAsNumber(weightLabel);
+
+				weight_value = ret_val;
+
```
howsohazard, line 630: Might as well remove the newline here to be consistent with the other code in this PR.

howsohazard: Same below -- looks like the newline phrasing style changed.

src/Amalgam/interpreter/Interpreter.cpp
```diff
@@ -873,6 +880,13 @@ bool Interpreter::PopulatePerformanceConstraintsFromParams(std::vector<Evaluable
 		}
 	}
 
+
+	//check if caller specifed override of the default warning collections behavior
+	if(params.size() > warning_override_offset)
```
howsohazard, line 885: To better match the documentation and simplify the code, the default behavior of collectWarnings to true should be based on if any_constraints is set rather than the length of parameters.  Though the warning_override_offset is still obviously needed for the InterpretNodeIntoBoolValue call.

src/Amalgam/interpreter/Interpreter.cpp
```diff
@@ -900,6 +914,8 @@ void Interpreter::PopulatePerformanceCounters(PerformanceConstraints *perf_const
 			perf_constraints->maxNumExecutionSteps = 1;
 			perf_constraints->curExecutionStep = 1;
 			perf_constraints->constraintsExceeded = true;
+
```
howsohazard: remove newline here

src/Amalgam/interpreter/Interpreter.cpp
```diff
@@ -920,6 +936,8 @@ void Interpreter::PopulatePerformanceCounters(PerformanceConstraints *perf_const
 		{
 			perf_constraints->maxNumAllocatedNodes = 1;
 			perf_constraints->constraintsExceeded = true;
+
```
howsohazard: remove newline

src/Amalgam/interpreter/Interpreter.h
```diff
@@ -25,6 +25,16 @@ class PerformanceConstraints
 {
 public:
 
+	enum class ViolationType { NoViolation, NodeAllocation, ExecutionStep, ExecutionDepth, ContainedEntitiesNumber, ContainedEntitiesDepth };
```
howsohazard: more than 120 chars, recommend breaking up

src/Amalgam/interpreter/Interpreter.h
```diff
@@ -25,6 +25,16 @@ class PerformanceConstraints
...
+
+	void AddWarning(std::string warning)
```
howsohazard: pass by reference to not make another extra copy of the string.  also would be worth a brief descriptive comment

src/Amalgam/interpreter/Interpreter.h
```diff
@@ -25,6 +25,16 @@ class PerformanceConstraints
 {
```
howsohazard, line 26: Given the expanded scope of this class, we discussed renaming to InterpreterConstraints or something similar/better.

src/Amalgam/interpreter/Interpreter.h
```diff
@@ -130,6 +140,24 @@ class PerformanceConstraints
 	//constrains the maximum length of an entity id (primarily to make sure it doesn't cause problems for file systems)
 	//If 0, then unlimited
 	size_t maxEntityIdLength;
+
+	//If true, collectWarnings
+	bool collectWarnings;
+
+	FastHashMap<std::string, size_t> &GetWarnings()
```
howsohazard: The method should be up with the other methods.

src/Amalgam/interpreter/Interpreter.h
```diff
@@ -130,6 +140,24 @@ class PerformanceConstraints
...
+
+	ViolationType constraintViolation;
+
+	private: 
+#ifdef MULTITHREAD_SUPPORT
+	Concurrency::ReadWriteMutex warningMutex;
+#endif
+
+	//maps warnings to the count of their occurence 
+	FastHashMap<std::string, size_t> warnings;
```
howsohazard: Having the getter for this above doesn't fit with how this class is used (more like a struct)

src/Amalgam/interpreter/Interpreter.h
```diff
@@ -993,6 +1021,8 @@ class Interpreter
 			if(performanceConstraints->curExecutionStep > performanceConstraints->maxNumExecutionSteps)
 			{
 				performanceConstraints->constraintsExceeded = true;
+
```
howsohazard: Newline is inconsistent with spacing below

src/Amalgam/interpreter/InterpreterOpcodesBase.cpp
```diff
@@ -562,10 +563,10 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_CALL_SANDBOXED(EvaluableNo
 	PopulatePerformanceCounters(perf_constraints_ptr, nullptr);
 
 	Interpreter sandbox(evaluableNodeManager, randomStream.CreateOtherStreamViaRand(),
-		writeListeners, printListener, perf_constraints_ptr, nullptr, this);
+		writeListeners, printListener, &perf_constraints, nullptr, this);
```
howsohazard: It looks like perf_constraints_ptr is no longer used in this method.  I'd suggest to remove, but it may be more consistent to change PopulatePerformanceConstraintsFromParams to return true if warnings are requested, and potentially change PopulatePerformanceConstraintsFromParams to reflect the corresponding name change of PerformanceConstraints

src/Amalgam/interpreter/InterpreterOpcodesBase.cpp
```diff
@@ -589,10 +590,72 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_CALL_SANDBOXED(EvaluableNo
...
+	if(perf_constraints.constraintsExceeded && perf_constraints.collectWarnings)
+		return BundleResultWithWarningsIfNeeded(EvaluableNodeReference::Null(), perf_constraints_ptr);
 
-	return result;
+	return BundleResultWithWarningsIfNeeded(result, perf_constraints_ptr);
+}
+
+static std::string ConstraintViolationToString(PerformanceConstraints::ViolationType violation)
+{
+
```
howsohazard: don't need extra newline here.  also, don't need a break; after returns below.

src/Amalgam/interpreter/InterpreterOpcodesBase.cpp
```diff
@@ -589,10 +590,72 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_CALL_SANDBOXED(EvaluableNo
...
+		//cases should be exhaustive, so this is unreachable
+		assert(false);
+		break;
+	}
+
+	assert(false);
+	return ""; //unreachable
+}
+
+const EvaluableNodeReference Interpreter::BundleResultWithWarningsIfNeeded(EvaluableNodeReference result, PerformanceConstraints *perf_constraints)
```
howsohazard: How is the const helping here?

src/Amalgam/interpreter/InterpreterOpcodesBase.cpp
```diff
@@ -589,10 +590,72 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_CALL_SANDBOXED(EvaluableNo
...
+{
+	if(perf_constraints == nullptr || !perf_constraints->collectWarnings)
+	{
+		return result;
+	}
+	
+	EvaluableNodeReference warning_assoc = CreateAssocOfNumbersFromIteratorAndFunctions(perf_constraints->GetWarnings(), [](std::pair<std::string, size_t> warning_count)
```
howsohazard: The documentation said this returns a list of warnings.  This looks like it returns an assoc with their counts.  Also, these lines are too long, I had to scroll quite a bit in the diff.

src/Amalgam/interpreter/InterpreterOpcodesBase.cpp
```diff
@@ -589,10 +590,72 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_CALL_SANDBOXED(EvaluableNo
...
+	EvaluableNodeReference result_tuple(evaluableNodeManager->AllocNode(ENT_LIST), true);
+
+	auto &result_tuple_ocn = result_tuple->GetOrderedChildNodesReference();
+
+	result_tuple_ocn.reserve(3);
+	result_tuple_ocn.push_back(result);
+	result_tuple_ocn.push_back(warning_assoc);
+	result_tuple_ocn.push_back(constraint_violation_string);
+
+	result_tuple.UpdatePropertiesBasedOnAttachedNode(result, false);
```
howsohazard: The booleans here are exactly the opposite of what they should be -- it should be true, false, false according to the documentation for UpdatePropertiesBasedOnAttachedNode.  However, assuming all of these results are unique, they don't make a difference to be specified, so you could ommit them.

src/Amalgam/interpreter/InterpreterOpcodesDataTypes.cpp
```diff
@@ -194,10 +195,29 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_SYMBOL(EvaluableNode *en,
...
+	// if didn't find it in the stack, try it in the labels
 	EntityReadReference cur_entity_ref(curEntity);
 	if(cur_entity_ref != nullptr)
-		return cur_entity_ref->GetValueAtLabel(sid, nullptr, true, true);
+	{
+		auto [value, found] = cur_entity_ref->GetValueAtLabel(sid, nullptr, true, true);
+
+		if(!found)
+		{
+			if(asset_manager.warnOnUndefined)
```
howsohazard: This logic should be an else under the performanceConstraints logic.  We don't want to print to stderr if the warnings are being captured.  Also worth renaming the performanceConstraints attribute (and variables) to reflect the updated name of PerformanceConstraints.

src/Amalgam/AmalgamMain.cpp
```diff
@@ -66,6 +66,8 @@ Usage: amalgam [options] [file]
 
     --debug-sources  When specified, prepends all node comments with the source of the node when applicable
 
+	--no-warn-on-undefined When specified, the amalgam will not emit warnings for undefined variables
```
howsohazard: Per the rest of the formatting here, and per the typical manpage layout, if the parameter is longer than the indentation, the text should start on the next line (see --p-count [number] above)

src/Amalgam/amlg_code/full_test.amlg
```diff
@@ -79,6 +79,8 @@
  (print (call_sandboxed (lambda (+ (+ y 4) 4)) (assoc y 3) (null) (null) 3) "\n")
  (print (call_sandboxed (lambda (+ (+ y 4) 4)) (assoc y 3) (null) (null) 1) "\n")
 
+ (print (call_sandboxed (lambda (+ (+ y 4) 4)) (null) (null) (null) 1 (false)) "\n")
```
howsohazard, line 82: worth adding code that tests the collection of warnings

src/Amalgam/interpreter/InterpreterOpcodesBase.cpp
naricc: Its still used; I have changed it so its passed here instead of taking the reference again.

Also did the name change.

src/Amalgam/interpreter/Interpreter.h
naricc: Deleted the method.

src/Amalgam/interpreter/Interpreter.h
naricc: Got rid of the getter.

src/Amalgam/interpreter/InterpreterOpcodesBase.cpp
naricc: Got rid of it.

src/Amalgam/interpreter/InterpreterOpcodesDataTypes.cpp
naricc: I'm a little confused by this comment; we don't have a performanceConstraints (it would be null) if we haven't been given any performance constraints. I can make these mutually exclusive by moving one into an else though.

src/Amalgam/amlg_code/full_test.amlg
naricc: Added.

src/Amalgam/interpreter/InterpreterOpcodesBase.cpp
naricc: Updated the documentation.

src/Amalgam/interpreter/Interpreter.h
naricc: I used rvalue reference.

src/Amalgam/interpreter/Interpreter.h
naricc: I've renamed it, and all the related variables named perf_constraints/perfConstraints/etc. that I could find.

src/Amalgam/interpreter/Interpreter.h
naricc: I put each field on its own line.

docs/language.js
```diff
@@ -30,7 +30,7 @@ var data = [
...
+		"description" : "String is parsed into code, and the result is returned.  If transactional is false, the default, it will attempt to parse the whole string and will return the closest code possible if there are any parse issues.  If transactional is true, it will parse the string transactionally, meaning that any node that has a parse error or is incomplete will be omitted along with all child nodes except for the top node.  If any performance constraints are given or return_warnings is true,  the result will be a tuple of the form [value, warnings, performance_constraint_violation], where warnings is an assoc mapping all undefined variable warnings to their number of occurrences, and perf_constraint violation is a string denoting the constraint exceeded (empty string if none), unless return_warnings is false, in which case just the value will be returned.",
```
howsohazard: All of the strings in the docs have an extra space there (which may render funny in the VSCode plugin)

docs/language.js
```diff
@@ -30,7 +30,7 @@ var data = [
...
 		"example" : "(parse \"(list 1 2 3 4 5)\")"
```
howsohazard, line 34: Also, maybe just read "all warnings" instead of "all undefined variable warnings", since we'll undoubtedly have more warnings in the future.

src/Amalgam/out.txt
```diff
@@ -284,13 +284,28 @@ if 2
...
+
+[11 {} ""]
+
+[(null) {} "Execution depth exceeded"]
+
 (null)
+
+[
+	(null)
+	{"x " 1 "y " 1}
```
howsohazard: These warnings don't look like readable warnings -- just seem that it's the variable with a space after it.  Worth digging into.

src/Amalgam/entity/EntityQueries.cpp
```diff
@@ -371,7 +371,18 @@ EvaluableNodeReference EntityQueryCondition::GetMatchingEntities(Entity *contain
...
+				double found;
+
+				if(e != nullptr)
+				{
+					std::tie(value, found) = e->GetValueAtLabelAsNumber(singleLabel);
+				}
+				else
+				{
+					found = false;
+				}
```
howsohazard: Don't need the {}'s around the if or else anymore.

src/Amalgam/entity/EntityQueries.cpp
```diff
@@ -629,13 +655,21 @@ EvaluableNodeReference EntityQueryCondition::GetMatchingEntities(Entity *contain
 			auto get_value = [matching_entities, this]
 			(size_t i, StringInternPool::StringID &value)
 			{
-				return matching_entities[i]->GetValueAtLabelAsStringId(singleLabel, value);
+				auto [ret_value, found] = matching_entities[i]->GetValueAtLabelAsStringId(singleLabel);
+
```
howsohazard, line 654: Newlines here and next couple blocks down don't match the rest of the code here in your PR.

src/Amalgam/interpreter/Interpreter.cpp
```diff
@@ -867,162 +874,174 @@ bool Interpreter::PopulatePerformanceConstraintsFromParams(std::vector<Evaluable
...
+				interpreter_constraints.maxEntityIdLength = static_cast<ExecutionCycleCount>(value);
 				any_constraints = true;
 			}
 		}
 	}
 
+
+	//check if caller specifed override of the default warning collections behavior
+	if(params.size() > warning_override_offset)
+		interpreter_constraints.collectWarnings = InterpretNodeIntoBoolValue(params[warning_override_offset]);
```
howsohazard: Need to pass the correct default into InterpretNodeIntoBoolValue, which should be any_constraints.  This is because if the user does specify this parameter but it's null, it should still use the default to be consistent.

src/Amalgam/interpreter/Interpreter.h
```diff
@@ -21,9 +21,27 @@
...
+	  ExecutionStep,
+	  ExecutionDepth,
+	  ContainedEntitiesNumber,
+	  ContainedEntitiesDepth
+  };
+
+  //Adds the string specified by warning to the list of warnings. Takes warning as an rvalue reference.
+  void AddWarning(std::string &&warning)
+  {
+#ifdef MULTITHREAD_SUPPORT
```
howsohazard: This and #endif should be one tab in (preprocessor macros one tab prior to code).

src/Amalgam/interpreter/Interpreter.h
```diff
@@ -156,7 +188,7 @@ class Interpreter
 	// if performance_constraints is not nullptr, then it will limit execution appropriately
```
howsohazard: Missed one! (and the parameter name below)

src/Amalgam/interpreter/InterpreterOpcodesBase.cpp
```diff
@@ -586,13 +587,67 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_CALL_SANDBOXED(EvaluableNo
...
+	}
+
+	assert(false);
+	return ""; //unreachable
+}
+
+EvaluableNodeReference Interpreter::BundleResultWithWarningsIfNeeded(EvaluableNodeReference result, InterpreterConstraints *interpreter_constraints)
+{
+	if(interpreter_constraints == nullptr || !interpreter_constraints->collectWarnings)
+	{
```
howsohazard: Don't need {}'s here

src/Amalgam/interpreter/InterpreterOpcodesDataTypes.cpp
howsohazard: The logic should be:
if we have interpreter constraints, then collect warnings
otherwise if we are printing warnings then print warnings

src/Amalgam/interpreter/Interpreter.cpp
howsohazard: Because of how this method is used elsewhere, should set any_constraints to true if collectWarnings is true.

src/Amalgam/interpreter/InterpreterOpcodesDataTypes.cpp
```diff
@@ -194,10 +195,29 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_SYMBOL(EvaluableNode *en,
...
+				if(interpreterConstraints != nullptr)
+					interpreterConstraints->AddWarning(sid->string + " " + en->GetCommentsString());
+			}
+		}
+
+		return value;
+	}
+
+	if(interpreterConstraints != nullptr && interpreterConstraints->collectWarnings)
+		interpreterConstraints->AddWarning(sid->string + " " + en->GetCommentsString());
```
howsohazard: Need to put the actual warning here and use logic (as is done elsewhere) whether to include the location or not if locations are being recorded.  E.g., if the en has a comment but locations are not being used, this should not be appended.

docs/language.js
naricc: You mean the extra one after the comma? 

There are two after the period, but I noticed this file is not entirely consistent about whether there are two or one spaces after a period.

src/Amalgam/out.txt
naricc: I've changed them as discussed.

src/Amalgam/interpreter/InterpreterOpcodesDataTypes.cpp
```diff
@@ -203,25 +204,33 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_SYMBOL(EvaluableNode *en,
...
-		interpreterConstraints->AddWarning(sid->string + " " + en->GetCommentsString());
-
+	EmitOrLogWarningIfNeeded(sid, en);
+	
 	return EvaluableNodeReference::Null();
 }
 
+void Interpreter::EmitOrLogWarningIfNeeded(StringInternPool::StringID sid, EvaluableNode *en)
+{
+	std::stringstream warning_stream;
```
howsohazard: Just use a std::string and concatenate.  std::stringstream comes with a lot of unnecessary overhead, and nothing is being used beyond concatenation.

src/Amalgam/interpreter/InterpreterOpcodesDataTypes.cpp
```diff
@@ -203,25 +204,33 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_SYMBOL(EvaluableNode *en,
...
+
+	warning_stream << "Warning: undefined symbol " << sid->string;
+
+	if(asset_manager.debugSources)
+		warning_stream << " " << en->GetCommentsString();
```
howsohazard: Should have logic to make sure that it has a non-empty comments string.  And if it does, then concatenate " at " followed by the subset of the comments string up to the newline.

src/Amalgam/interpreter/InterpreterOpcodesDataTypes.cpp
```diff
@@ -203,25 +204,33 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_SYMBOL(EvaluableNode *en,
...
+
+	if(asset_manager.warnOnUndefined)
```
howsohazard: This should be under the else of collecting warnings below, because we do not want to emit this to stderr if warnings are being collected.

howsohazard: Though in general, call_sandboxed should never emit warnings, so I think it should be based on the presence of interpreterConstraints being not nullptr only (regardless of collectWarnings).

src/Amalgam/entity/Entity.h
```diff
@@ -295,11 +295,11 @@ class Entity
...
 		bool on_self = false, bool batch_call = false);
```
howsohazard, line 299: There's an extra space between > and GetValueAtLabel

src/Amalgam/entity/EntityQueries.cpp
```diff
@@ -600,17 +616,23 @@ EvaluableNodeReference EntityQueryCondition::GetMatchingEntities(Entity *contain
...
-				return matching_entities[i]->GetValueAtLabelAsNumber(weightLabel, weight_value);
+				auto [ret_val, found] = matching_entities[i]->GetValueAtLabelAsNumber(weightLabel);
+				weight_value = ret_val;
+
+				return found;
 			};
 
 			auto value_weights = EntityQueriesStatistics::ValueMassesNumber<size_t>(0, matching_entities.size(), matching_entities.size(),
-				get_value, weightLabel != StringInternPool::NOT_A_STRING_ID, get_weight);
+																					get_value, weightLabel != StringInternPool::NOT_A_STRING_ID, get_weight);
```
howsohazard, line 635: Was this extra newline intended?

src/Amalgam/interpreter/Interpreter.cpp
```diff
@@ -867,162 +874,179 @@ bool Interpreter::PopulatePerformanceConstraintsFromParams(std::vector<Evaluable
...
 			}
 		}
 	}
 
+
+	//check if caller specifed override of the default warning collections behavior
+	if(params.size() > warning_override_offset)
+	{
+		interpreter_constraints.collectWarnings = InterpretNodeIntoBoolValue(params[warning_override_offset], any_constraints);
+		any_constraints = true;
```
howsohazard: any_constraints should only be set to true if collectWarnings is true

src/Amalgam/interpreter/Interpreter.h
```diff
@@ -21,9 +21,27 @@
 class EntityQueryCondition;
 
 //Manages performance constraints and accompanying performance counters
-class PerformanceConstraints
+class InterpreterConstraints
 {
 public:
+  enum class ViolationType
```
howsohazard: Indentation of this enum class and the comment below it aren't correct

src/Amalgam/interpreter/Interpreter.h
```diff
@@ -130,6 +148,20 @@ class PerformanceConstraints
...
+
+	//If true, collectWarnings
+	bool collectWarnings;
+
+	ViolationType constraintViolation;
+
+	//maps warnings to the count of their occurrence 
+	FastHashMap<std::string, size_t> warnings;
+
+	private:
```
howsohazard: private shouldn't be indented

src/Amalgam/interpreter/Interpreter.h
```diff
@@ -1272,11 +1311,13 @@ class Interpreter
...
 
 	//if not nullptr, then contains the respective constraints on performance
-	PerformanceConstraints *performanceConstraints;
+	InterpreterConstraints *interpreterConstraints;
 
 	//a stack (list) of the current nodes being executed
 	std::vector<EvaluableNode *> *opcodeStackNodes;
 
+
+
```
howsohazard: extra newline

src/Amalgam/interpreter/InterpreterOpcodesBase.cpp
```diff
@@ -586,13 +587,65 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_CALL_SANDBOXED(EvaluableNo
...
+EvaluableNodeReference Interpreter::BundleResultWithWarningsIfNeeded(EvaluableNodeReference result, InterpreterConstraints *interpreter_constraints)
+{
+	if(interpreter_constraints == nullptr || !interpreter_constraints->collectWarnings)
+		return result;
+	
+	EvaluableNodeReference warning_assoc = CreateAssocOfNumbersFromIteratorAndFunctions(interpreter_constraints->warnings, [](std::pair<std::string, size_t> warning_count)
+																						{ return string_intern_pool.CreateStringReference(warning_count.first); }, [](std::pair<std::string, size_t> warning_count)
+																						{ return static_cast<double>(warning_count.second); }, evaluableNodeManager);
+
+	EvaluableNodeReference constraint_violation_string = EvaluableNodeReference(evaluableNodeManager->AllocNode(ConstraintViolationToString(interpreter_constraints->constraintViolation)), true);
```
howsohazard: As we discussed, this should be nullptr if there's no constraint rather than empty string.  Will need to update documentation as well.

src/Amalgam/interpreter/InterpreterOpcodesDataTypes.cpp
howsohazard: Make sure to remove the include above too.

src/Amalgam/interpreter/InterpreterOpcodesDataTypes.cpp
```diff
@@ -194,14 +196,41 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_SYMBOL(EvaluableNode *en,
...
-	//if didn't find it in the stack, try it in the labels
+	// if didn't find it in the stack, try it in the labels
 	EntityReadReference cur_entity_ref(curEntity);
 	if(cur_entity_ref != nullptr)
-		return cur_entity_ref->GetValueAtLabel(sid, nullptr, true, true);
+	{
+		auto [value, found] = cur_entity_ref->GetValueAtLabel(sid, nullptr, true, true);
+
+		if(!found)
+		{
```
howsohazard: Don't need {}'s for one-line if.

src/Amalgam/interpreter/InterpreterOpcodesBase.cpp
naricc: Fixed and updated the documentation.

src/Amalgam/interpreter/Interpreter.cpp
```diff
@@ -885,7 +885,7 @@ bool Interpreter::PopulateInterpreterConstraintsFromParams(std::vector<Evaluable
 	if(params.size() > warning_override_offset)
 	{
 		interpreter_constraints.collectWarnings = InterpretNodeIntoBoolValue(params[warning_override_offset], any_constraints);
-		any_constraints = true;
+		any_constraints = interpreter_constraints.collectWarnings;
 	}
```
howsohazard, line 889: Needs to be:
if(interpreter_constraints.collectWarnings)
  any_constraints = true;
Because otherwise if collectWarnings is false, it'll say there are none.

src/Amalgam/interpreter/InterpreterOpcodesDataTypes.cpp
```diff
@@ -203,32 +202,35 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_SYMBOL(EvaluableNode *en,
...
+	warning.append("Warning: undefined symbol " + sid->string);
 
 	if(asset_manager.debugSources)
-		warning_stream << " " << en->GetCommentsString();
+		warning.append( " " + en->GetCommentsString());
 
-	if(asset_manager.warnOnUndefined)
-		std::cerr << warning_stream.str() << std::endl;
 
 	if(interpreterConstraints != nullptr && interpreterConstraints->collectWarnings)
```
howsohazard: Need to remove the interpreterConstraints->collectWarnings out of the first if

docs/language.js
howsohazard: No, there's an extra space between "true,  the"

src/Amalgam/interpreter/InterpreterOpcodesDataTypes.cpp
```diff
@@ -194,14 +195,44 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_SYMBOL(EvaluableNode *en,
...
 
+void Interpreter::EmitOrLogWarningIfNeeded(StringInternPool::StringID sid, EvaluableNode *en)
+{
+	std::string warning = "";
+
+	warning.append("Warning: undefined symbol " + sid->string);
+
+	if(asset_manager.debugSources)
+		warning.append( " " + en->GetCommentsString());
+
```
howsohazard, line 231: Extra newline here.  Didn't address my previous comment recommendation on how to handle this.

naricc: I'm confused what isn't addressed here? Just use append instead of +?

naricc: Ok, I figured out what was meant here.

src/Amalgam/interpreter/Interpreter.h
```diff
@@ -982,43 +1014,50 @@ class Interpreter
...
 				return true;
 			}
 		}
 
 		//return whether they have ever been exceeded
-		return performanceConstraints->constraintsExceeded;
+		return interpreterConstraints->constraintsExceeded;
 	}
+	
+	EvaluableNodeReference BundleResultWithWarningsIfNeeded(EvaluableNodeReference result, InterpreterConstraints *interpreter_constraints_ptr);
```
howsohazard: Need to comment on what these methods are for, this and EmitOrLogWarningIfNeeded

howsohazard: In particular, I just noticed that EmitOrLogWarningIfNeeded only emits symbol not found warnings.  So either it should be renamed for the warning it emits (and rename sid to variable_not_found_sid or something similar) or it should be genericized such that it can emit any kind of warning and the warning string passed in as the parameter (in which case the method can optionally append the location and then store it where appropriate).

naricc: I decided to just rename it since its not exactly clear yet how the genericized version would be used. I also added doc strings and changed the name of the parameter "sid".

howsohazard (approved):

PR #339, opened 2025-01-08
22484: Fix assign_entity_roots transaction logs.

dmaze:
If any form of transaction logging is turned on, wrap the node of the assign_entity_roots call in a lambda so that labels don't get lost.

In the EvalEntityJsonPtrLogged() API call, make sure to emit labels in the log entry that comes back.

### DIFF
--- a/test/lib_smoke_test/main.cpp
+++ b/test/lib_smoke_test/main.cpp
@@ -140,7 +140,9 @@ static void loadAndEval(TestResult &test_result)
 }
 
 static std::string handle("handle");
+static std::string handle2("handle2");
 static std::string filename("counter.amlg");
+static std::string filename2("counter2.amlg");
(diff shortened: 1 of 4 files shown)


test/lib_smoke_test/main.cpp
```diff
@@ -258,6 +260,51 @@ static void executeEntityJsonLoggedTwice(TestResult &test_result) {
 	}
 }
 
+static void executeCounter2(TestResult &test_result) {
```
howsohazard: For style consistency, the {'s should be on new lines.

howsohazard (approved):

test/lib_smoke_test/main.cpp
dmaze: Fixed.

PR #342, opened 2025-01-12
22518: Improves memory use for current_index via better use of immediate values, cleans up some garbage collection, corrects IDE style settings

howsohazard:


src/Amalgam/Parser.cpp
```diff
@@ -164,6 +164,19 @@ EvaluableNodeReference Parser::ParseFromKeyStringId(StringInternPool::StringID c
...
+StringInternPool::StringID Parser::ParseFromKeyStringIdToStringIdWithReference(StringInternPool::StringID code_string_id)
+{
+	if(code_string_id == string_intern_pool.NOT_A_STRING_ID)
+		return string_intern_pool.NOT_A_STRING_ID;
+
+	std::string &code_string = code_string_id->string;
+	if(code_string.size() == 0 || code_string[0] != '\0')
+		return string_intern_pool.CreateStringReference(code_string_id);
+
+	std::string escaped_string(&code_string[1], code_string.size() - 1);
```
naricc, line 176: I am a little confused by the naming here, since I don't see anything going on that looks like "escaping" the string. But it does seem consistent with the naming in other methods.

naricc:
There seem to be two independent things here; the change to the heap limit that triggers GC, and the use of a better allocation path in`InterpretNode_ENT_CURRENT_INDEX` for immediate use. 

Do we have any indication which if these results in the performance boost?

howsoRes (approved):

src/Amalgam/Parser.cpp
howsohazard: Key string ids can have escaping for code-as-keys.  So here, if it is code, it just re-stringifies it by taking the later part.

PR #348, opened 2025-02-07
22772: Reduces logic overhead in garbage collection slightly

howsohazard:


src/Amalgam/evaluablenode/EvaluableNodeManagement.cpp
```diff
@@ -893,59 +896,95 @@ std::pair<bool, bool> EvaluableNodeManager::UpdateFlagsForNodeTreeRecurse(Evalua
...
+		return;
+
 #ifdef AMALGAM_FAST_MEMORY_INTEGRITY
 	assert(tree->IsNodeValid());
 #endif
 
-	//if entering this function, then the node hasn't been marked yet
-	tree->SetKnownToBeInUseAtomic(true);
+	tree->SetKnownToBeInUse(true);
+	//because marking occurs during garbage collection and threadLocalAllocationBuffer
```
naricc, line 953: I think we should add a comment by the declaration of the threadLocalAllocationBuffer itself that it is also being used for this purpose. If some one makes changes to how it works in the future, its going to be pretty easy to overlook this.

naricc (approved):

howsoRes (approved):

PR #352, opened 2025-02-11
22800: Fixes an issue where the absolute value of (null) returns zero

cademack:


src/Amalgam/interpreter/Interpreter.h
```diff
@@ -515,7 +515,7 @@ class Interpreter
 		return InterpretNode(n, immediate_result);
 	}
 
-	//computes a unary numeric function on the given node
+	//computes a unary numeric function on the given node, returns a null ENR if n interprets as (null)
```
howsohazard: Change comment to be ", returns an ENT_NULL if n is interpreted as an ENT_NULL

howsohazard (approved):

PR #356, opened 2025-02-14
13225: Implements ability to dynamically select number of entities to return for distance queries, MINOR

howsohazard:


src/Amalgam/entity/EntityQueriesStatistics.h
```diff
@@ -723,164 +725,310 @@ class EntityQueriesStatistics
...
-		//has_weight, if set, will use get_weight, taking in a function of an entity reference and a reference to an output double to set the weight,
-		// and should return true if the entity has a weight, false if not
-		//sort_results, if set, will sort the results appropriately for the distance_weight_exponent,
-		// from smallest to largest if distance_weight_exponent is positive, largest to smallest otherwise
-		//get_entity returns the EntityReference for an iterator of EntityDistancePairContainer
-		//get_distance_ref returns a reference as a pointer to the location of the distance in the EntityDistancePairContainer
-		inline void TransformDistances(std::vector<DistanceReferencePair<EntityReference>> &entity_distance_pair_container, bool sort_results)
+		__forceinline static double ConvertSurprisalToProbability(double surprisal, double weight)
+		{
+			double prob = ConvertSurprisalToProbability(surprisal);
```
howsoRes: for performance, does it make sense to add an `if` that checks if weight=1? ie if weight==1 return ConvertSuprisalToProbability else run the method as is?

howsoRes: or if this method is for those places that only specifically have weights, it's okay as-is, question is - where is this method still being used? I'm not sure it is, so you can probably remove it if so

src/Amalgam/entity/EntityQueriesStatistics.h
```diff
@@ -723,164 +725,310 @@ class EntityQueriesStatistics
...
+								return std::make_tuple(1.0, 1.0, 1.0, 1.0);
+
+							double weight = 1.0;
+							getEntityWeightFunction(iter->reference, weight);
+
+							return std::make_tuple(weight, 1.0, weight, weight);
+						}, result_func);
 				}
-				else if(distanceWeightExponent != 1)
+				else if(distanceWeightExponent == 1)
```
howsoRes: should this block be above the previous one, since dt=-1 and dt=1 would be the most common use cases and thus we'd want them to be hit first in this else-if flow?

src/Amalgam/entity/EntityQueryBuilder.h
```diff
@@ -336,19 +336,50 @@ namespace EntityQueryBuilder
...
+				//retrieve all the parameters from the list, clamping as appropriate
+				if(num_params >= 1)
+				{
+					double min_inc_prob = EvaluableNode::ToNumber(top_k_ocn[0], 0.0);
+					cur_condition->numToRetrieveMinIncrementalProbability = std::max(0.0, min_inc_prob);
+
+					if(num_params >= 2)
+					{
+						double min_to_retrieve = EvaluableNode::ToNumber(top_k_ocn[1], std::numeric_limits<double>::infinity());
+						min_to_retrieve = std::max(0.0, min_to_retrieve);
```
howsoRes, line 364: probably should be std::max(1.0, min_to_retrieve), right? not sure that it make sense if user passed in a 0 as the second parameter? same for max_to_retrieve, shouldn't it be clamped to 1.0 as well? that would match the 'single value for k' logic which sets it to 1

src/Amalgam/entity/EntityQueriesStatistics.h
howsohazard: This method is no longer used, so just removing it.

src/Amalgam/entity/EntityQueriesStatistics.h
howsohazard: sure, done

src/Amalgam/entity/EntityQueryBuilder.h
howsohazard: I could foresee use cases where nothing is greater than 0.05 and so you don't have supporting data.  Not that we should do that now, but no reason to require one case, even if that's the way it works now.  I think leaving it in is less hidden wrt logic.

howsoRes (approved):

src/Amalgam/entity/EntityQueriesStatistics.h
```diff
@@ -723,164 +725,304 @@ class EntityQueriesStatistics
...
-				else if(distanceWeightExponent == 0)
+				else if(distanceWeightExponent == 1)
 				{
-					for(auto iter = begin(entity_distance_pair_container); iter != end(entity_distance_pair_container); ++iter)
-						iter->distance = 1.0;
+					num_to_keep = SelectBandwidthFromDistanceTransforms(
+						entity_distance_pair_container_begin, entity_distance_pair_container_end,
+						[this](auto iter)
+					{
+						double prob = 1.0 / iter->distance;
```
cademack, line 926: looks like this still works as if DWE is -1? but this is the DWE=1 path

src/Amalgam/entity/EntityQueryCaches.cpp
```diff
@@ -320,8 +326,9 @@ void EntityQueryCaches::GetMatchingEntities(EntityQueryCondition *cond, BitArray
 				}
 				else if(cond->queryType == ENT_QUERY_NEAREST_GENERALIZED_DISTANCE)
 				{
+
```
cademack: unintentional new line?

howsohazard: yep

PR #368, opened 2025-03-07
22953: Consolidates code for nearest neighbors, improves performance of many workflows

howsohazard:


src/Amalgam/SeparableBoxFilterDataStore.h
```diff
@@ -492,22 +492,80 @@ class SeparableBoxFilterDataStore
...
+		possible_knn_indices = enabled_indices;
+
+		//remove search_index so it doesn't find itself
+		possible_knn_indices.erase(search_index);
+
+		if(expand_to_first_nonzero_distance)
+			FindNearestEntities<true>(r_dist_eval, position_label_sids, top_k,
+				radius_label, possible_knn_indices, distances_out, ignore_index, rand_stream);
+		else
+			FindNearestEntities<true>(r_dist_eval, position_label_sids, top_k,
```
howsoRes: should be `<false>`   here, right?

src/Amalgam/SeparableBoxFilterDataStore.cpp
```diff
@@ -525,229 +498,91 @@ void SeparableBoxFilterDataStore::FindEntitiesNearestToIndexedEntity(Generalized
...
-		//skip this entity in the next loops
-		possible_knn_indices.erase(entity_index);
-		
-		potential_good_matches.pop();
 	}
 
 	//if we did not find K results (search failed), we must populate the remaining K cases/results to search from another way
 	//we will randomly select additional nodes to fill K results.  random to prevent bias/patterns
-	while(sorted_results.Size() < top_k && possible_knn_indices.size() > 0)
+	while(sorted_results.Size() < top_k)
```
howsoRes, line 529: could this get stuck in an inf loop if top_k > enabled_indices size? why did you remove that && ?

howsohazard: no because of line 486

howsoRes (approved):

PR #376, opened 2025-04-02
23110: Greatly improves performance of conviction queries on multithreaded machines, small performance improvements to distance queries, cleans up code

howsohazard:


src/Amalgam/Concurrency.h
```diff
@@ -87,3 +91,45 @@ namespace Concurrency
...
+inline void IterateOverConcurrentlyIfPossible(ContainerType &container, FunctionType func,
+	bool run_concurrently = false, bool urgent = false)
+{
+	size_t index = 0;
+#ifdef MULTITHREAD_SUPPORT
+	if(run_concurrently && container.size() > 1)
+	{
+		auto &thread_pool = (urgent ? Concurrency::urgentThreadPool : Concurrency::threadPool);
+		auto enqueue_task_lock = thread_pool.AcquireTaskLock();
+		if(thread_pool.AreThreadsAvailable())
```
cademack, line 107: So nothing happens if no threads are available?

cademack: well - I suppose it could fall through to the non-concurrent logic?

cademack (approved):

src/Amalgam/Concurrency.h
howsohazard: that's exactly the intent

PR #381, opened 2025-04-15
23430: Fixes rare segfault when multithreading creates and deletes nearly unique (one or two references) strings concurrently at high throughput

howsohazard:


src/Amalgam/string/StringInternPool.h
```diff
@@ -194,96 +194,51 @@ class StringInternPool
 		ValidateStringIdExistence(id);
 	#endif
 
-		int64_t refcount = id->refCount--;
-
-		//if other references, then can't clear it; signed, so it won't wrap around
```
dmaze, line 199: (I'd probably try to keep some of these comments.)

src/Amalgam/string/StringInternPool.h
```diff
@@ -194,96 +194,51 @@ class StringInternPool
...
+			if(std::atomic_compare_exchange_weak(&id->refCount, &ref_count, ref_count - 1))
+				return;
+		}
 
-		Concurrency::SingleLock lock(mutex);
+		Concurrency::Lock lock(mutex);
 
-		//with the lock, decrement reference count in case this string should stay active
-		refcount = id->refCount--;
+		//retry decrementing in case it has been modified by another thread
```
dmaze: So the possible cases here are:

1. `refCount` was previously at least 2; so we decremented the reference count and returned.
2. `refCount` was previously at most 1, but something else has incremented it.  We need to decrement the reference count and return.
3. `refCount` was previously at most 1, but something else concurrently called this method and it's gone.  `FastHashMap::erase()` seems to be idempotent so there's not actually a problem here.
4. `refCount` was previously at most 1, and it still has the same value.  This probably means nothing else has happened and we should move forward.

At this point I might check to see if `id->refCount.load() == ref_count` still, and if it isn't, repeat the initial loop (releasing the lock), which might simplify the logic a little bit.

But since `id->refCount` can change independently of the mutex, you're always going to have a race condition here unless you put a lock in `CreateStringReference(StringID)` and similar places (at which point the reference count is protected by the lock and doesn't need to be atomic).  I'm not sure a second check here adds value since the reference count still gets to change after this second check but before you've called `erase()`.

src/Amalgam/string/StringInternPool.h
```diff
@@ -194,96 +194,51 @@ class StringInternPool
...
-
-		//need to remove at least one reference, so put all counts back while wait for lock
-		for(auto r : references_container)
-		{
-			StringID id = get_string_id(r);
-			if(id != NOT_A_STRING_ID && id != emptyStringId)
-				id->refCount++;
-		}
-
-		Concurrency::SingleLock lock(mutex);
```
dmaze, line 267: Noting that the previous version of this gets the lock only once, regardless of how many IDs are deleted; this change gets the lock once per ID.  I don't know if that's a significant performance cost.

howsohazard: There are pros and cons to this.  However, attempting to lock all at once when you don't want to decrement and reincrement is far more expensive than the gains we would get from putting it all under one lock in this new technique.

src/Amalgam/string/StringInternPool.h
howsohazard: The only way refCount can change independently is to go up, which will prevent the deletion from occurring.  The purpose of this second check under the lock is to ensure that it can only go up, as the value can go down (but not below 1) outside of the lock.

dmaze (approved):

PR #400, opened 2025-05-29
23816: Fixes crash if multiple calls to SetMaxNumActiveThreads are made concurrently

howsohazard:


mkoistinen (approved):
LGTM! Thanks for the fast turnaround!

PR #456, opened 2025-10-29
24630: load/store to/from memory, entities, MAJOR

dmaze:
Add new LoadEntityFromMemory()/StoreEntityToMemory() external API calls that take the entity contents from a pair of void *, size_t rather than a file.

For all of the Load/StoreEntity APIs, add a pair of char **, size_t parameters that identify a path to a non-root entity.  This allows loading or storing something nested potentially deep within the entity tree.

Existing calls to LoadEntity() and StoreEntity() need to add the additional entity-path parameters, which will usually be nullptr, 0.  If LoadEntity is passed a non-null entity path, then its return value will include an entity path as well, which needs to be separately freed.

### DIFF
--- a/TRACE-FILES.md
+++ b/TRACE-FILES.md
@@ -1,0 +1,138 @@
+# Trace Files
+
+The Amalgam runtime uses a simple line-oriented format to record and replay API operations.  If you run the Amalgam CLI tool like `amalgam-st --trace`, you can type these commands directly into the tool to run top-level Amalgam API operations without necessarily writing a C program.
+
+The [Amalgam Python client library](https://github.com/howsoai/amalgam-lang-py) can generate trace files if an option `trace=True` is passed to the `Amalgam()` constructor.  The Amalgam CLI tool can consume these trace files via the `amalgam-st --tracefile [file]` option.  Trace files can be hand-written if needed.
+
+## Syntax
+
+Blank lines and lines that begin with a comment marker, `#` followed by at least one space, are ignored.  Lines with unknown commands or syntactic errors are ignored, and do not stop execution.
+
+Command lines contain a command name followed by some number of space-separated arguments.  An argument may appear in double quotes; if it does, the argument continues through the next matching double-quote.  If a double-quote is needed in an argument value, it may be preceded by a backslash.  No other escaping is possible or allowed, and backslashes that are not followed by double quotes are literally read as backslashes.  If ASCII double-quote bytes 0x22 are escaped by preceding them with a backslash byte 0x5C then an item can otherwise contain arbitrary binary data.
+
+Some commands end with a JSON object.  This is passed directly as a string, and is not quoted or escaped in any way.  Other commands can take a JSON object as a parameter, and this must be quoted and escaped as described above.  Compare:
+
+```none
+LOAD_ENTITY "handle" "path" "" false "{\"executeOnLoad\": true}"
+SET_ENTITY_PERMISSIONS "handle" {"load": true, "store": true}
+```
+
+Several of the entity-related operations take a `persistent` parameter.  This is a Boolean value, true if the string is exactly `true` and false otherwise.  This is conventionally not quoted but double quotes will be recognized and removed.
+
+## Commands
+
+The commands mirror the C API in `Amalgam.h`.  Commands are as follows:
+
+```none
+LOAD_ENTITY "handle" "path" "file_type" persistent "json_payload" "transaction_listener_path" "print_listener_path" "rand_seed" "entity_path"
+```
+
+Load an entity from a file on disk.  `handle` and `path` are required, all other parameters are optional.  If `entity_path` is provided then it contains a space-separated ordered list of entity IDs, and the entity is loaded into an interior entity underneath an existing `handle`.
+
+```none
+LOAD_ENTITY_FROM_MEMORY "handle" "base64_data" "file_type" "persistent" "json_payload" "transaction_listener_path" "print_listener_path" "rand_seed" "entity_path"
+```
+
+Load an entity from a base64-encoded data string.  `handle`, `base64_data`, and `file_type` are all required.  Other parameters are handled as `LOAD_ENTITY`.
+
+```none
+GET_ENTITY_PERMISSIONS "handle"
+```
+
+Get the current permissions of the specified entity.  `handle` is required.
+
+```none
+SET_ENTITY_PERMISSIONS "handle" json
+```
+
+Set the permissions of an entity.  `handle` is required, and is followed by an unquoted JSON string with the actual permissions.
+
+```none
+CLONE_ENTITY "from_handle" "to_handle" "file_type" persistent "json_payload" "transaction_listener_path" "print_listener_path"
+```
+
+Make a copy of an existing entity.  The two handles are required.
+
+```none
+STORE_ENTITY "handle" "path" "file_type" persistent "json_payload" "entity_path"
+```
+
+Store an entity to a file on disk.  `handle` and `path` are required, other options default to the options provided when the entity was created.
+
+```none
+STORE_ENTITY_TO_MEMORY "handle" "file_type" persistent "json_payload" "entity_path"
+```
+
+Store an entity to memory.  `handle` is required, and the output of the trace operation is the base64-encoded entity data.
+
+```none
+DESTROY_ENTITY "handle"
+```
+
+Destroy an entity.  `handle` is required.
+
+```none
+SET_JSON_TO_LABEL "handle" "label" json
+```
+
+Set the value at some label within an existing entity.  `handle` and `label` are required, and are followed by an unquoted JSON string with the content to set.
+
+```none
+GET_JSON_FROM_LABEL "handle" "label"
+```
+
+Get the value at some label within an existing entity.  `handle` and `label` are required, and the output of the trace operation is the JSON-encoded value.
+
+```none
+EXECUTE_ENTITY_JSON "handle" "label" json
+```
+
+Execute some label within an existing entity, passing a JSON object as context.  `handle` and `label` are required, and are followed by an unquoted JSON string with values to set in the execution context.  If the JSON string is empty or absent, it is treated as an empty object `{}`.  The executing Amalgam code will see the keys and values of the JSON object as ordinary variables.  The output of the trace operation is the JSON-encoded result of executing the label.
+
+```none
+EXECUTE_ENTITY_JSON_LOGGED "handle" "label" json
+```
+
+As `EXECUTE_ENTITY_JSON`, but also writes out an Amalgam-syntax description of how the execution modified the entity.
+
+```none
+EVAL_ON_ENTITY "handle" "amlg"
+```
+
+Execute arbitrary Amalgam code in the context of an existing entity.  `handle` and `amlg` are required, and `amlg` must be correctly quoted.  The output of the trace operation is the JSON-encoded result of evaluating the expression.
+
+```none
+SET_RANDOM_SEED "handle" content
+```
+
+Sets the random seed of an existing entity to arbitrary binary data.  `handle` is required, and `content` are any bytes remaining after any whitespace through the end of the line.
+
+```none
+VERSION
+```
+
+The output of the trace operation is the current Amalgam library version.  Takes no parameters.
+
+```none
+VERIFY_ENTITY "path"
+```
+
+Check that a filesystem path can be loaded as an entity.  `path` is required.
+
+```none
+GET_MAX_NUM_THREADS
+```
+
+The output of the trace operation is the number of concurrent threads the Amalgam interpreter can use.  Error on non-multithreaded Amalgam runtimes such as `amalgam-st`.
+
+```none
+SET_MAX_NUM_THREADS n
+```
+
+Set the maximum number of threads in a multithreaded Amalgam runtime.  The parameter is an unquoted number with the number of threads, and must be greater than zero.  Error on non-multithreaded Amalgam runtimes such as `amalgam-st`.
+
+```none
+EXIT
+```
+
+End the trace file.
--- a/src/Amalgam/Amalgam.h
+++ b/src/Amalgam/Amalgam.h
@@ -22,6 +22,11 @@ extern "C"
 		bool loaded;
 		char *message;
 		char *version;
+		//If an entity_path parameter was passed to LoadEntity, the path where the entity was
+		//actually loaded.  If non-null, contains entity_path_len entries.  Both the entries
+		//and the array itself need to be freed via DeleteString.
+		char **entity_path;
+		size_t entity_path_len;
 	};
 
 	//output from ExecuteEntityJsonPtrLogged
@@ -33,7 +38,13 @@ extern "C"
 
 	//loads the entity specified into handle
 	AMALGAM_EXPORT LoadEntityStatus LoadEntity(char *handle, char *path, char *file_type,
-		bool persistent, char *json_file_params, char *write_log_filename, char *print_log_filename);
+		bool persistent, char *json_file_params, char *write_log_filename, char *print_log_filename,
+		const char **entity_path, size_t entity_path_len);
+
+	//loads the entity specified into handle, from a memory buffer
+	AMALGAM_EXPORT LoadEntityStatus LoadEntityFromMemory(char *handle, void *data, size_t len, char *file_type,
+		bool persistent, char *json_file_params, char *write_log_filename, char *print_log_filename,
+		const char **entity_path, size_t entity_path_len);
 
 	//verifies the entity specified by path. Uses LoadEntityStatus to return any errors and version
 	AMALGAM_EXPORT LoadEntityStatus VerifyEntity(char *path);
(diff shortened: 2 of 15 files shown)


src/Amalgam/entity/EntityWriteListener.cpp
```diff
@@ -15,16 +15,15 @@ EntityWriteListener::EntityWriteListener(Entity *listening_entity,
 	fileSuffix = ")\r\n";
 	pretty = _pretty;
 	sortKeys = sort_keys;
-	if(!filename.empty())
+	if(logFile && logFile->good())
```
howsohazard: Even though this is correct and works as intended, and is the style in some places, for reading clarity it's better to check if it's null explicitly.  The compiler will optimize it out.

dmaze: Updated these (with a detour noticing that `std::unique_ptr<T>::operator ==(std::nullptr_t)` is a special case).

src/Amalgam/AssetManager.cpp
```diff
@@ -119,6 +120,38 @@ void AssetManager::AssetParameters::UpdateResources()
 	}
 }
 
+std::unique_ptr<std::istream> AssetManager::AssetParameters::OpenRead()
+{
+	if(toMemory)
+	{
```
howsohazard: In the style we use, don't use {}'s if only one line after if.

src/Amalgam/AssetManager.cpp
```diff
@@ -119,6 +120,38 @@ void AssetManager::AssetParameters::UpdateResources()
...
+		return std::make_unique<std::istringstream>(resourceContents);
+	}
+	return std::make_unique<std::ifstream>(resourcePath, std::ios::binary | std::ios::in);
+}
+
+// String-output stream that records the final result in an asset parameters object.
+class AssetParametersStringStream : public std::ostringstream
+{
+	AssetManager::AssetParameters *asset_parameters;
```
howsohazard: Attributes go after methods in our style.

src/Amalgam/AssetManager.cpp
```diff
@@ -119,6 +120,38 @@ void AssetManager::AssetParameters::UpdateResources()
...
+
+public:
+	AssetParametersStringStream(AssetManager::AssetParameters *asset_parameters) : std::ostringstream(), asset_parameters(asset_parameters)
+	{}
+	virtual ~AssetParametersStringStream()
+	{
+		asset_parameters->resourceContents = str();
```
howsohazard: it seems weird that the destructor modifies asset_params; I don't fully understand why

src/Amalgam/AssetManager.cpp
howsohazard: same below for the other if(toMemory)

src/Amalgam/AssetManager.h
```diff
@@ -119,6 +121,12 @@ class AssetManager
 		//updates resources based on the parameters -- should be called after SetParams
 		void UpdateResources();
 
+		//open the resource or content for reading
+		std::unique_ptr<std::istream> OpenRead();
```
howsohazard: Is it weird these don't take any params?  or should comment that they're based on the attributes.  Could these be templatized or something for read/write?  OpenResource<type>?

src/Amalgam/AssetManager.cpp
dmaze: The problem this class is trying to solve is that `AssetParameters::OpenWrite` returns a generic `std::ostream`, but in particular we need a stream that captures what's written to it, probably in the `AssetParameters`, without the caller really being aware of this.

So the flow here is

1. `AssetManager::StoreResource()` calls `asset_params->OpenWrite()`, and gets back a stream (which could be either file or string-backed)
2. `StoreResource()` writes whatever data to the stream
3. The `std::unique_ptr<str::ostream>` goes out of scope, which triggers its destructor.
4. If you were writing to memory, this code runs, captures the stream's contents, and records them in the asset parameters to return to the original caller.

As I'm writing this, and matching it up with the code, it's a little overengineered: `AssetParameters::OpenWrite()` is _only_ called from `StoreResource()`, so if I refactor that to `StoreResourceToStream(std::ostream &)` and do the stream setup around that, I can get rid of the `OpenWrite()` mechanism and this class.  Similarly `OpenRead()` is only called twice, on two different paths that write caml files (everything else has a faster path to use the string directly).

So...I think it does make sense to get rid of this class and this mechanism, I'll do that.

src/Amalgam/AssetManager.h
dmaze: These methods are gone.

src/Amalgam/AssetManager.cpp
dmaze: The only place this happened was in this class, which has gone away now.

src/Amalgam/AssetManager.cpp
dmaze: Fixed all of these.

(I am very used to style rules where you _must_ use the braces, to avoid bugs where you add a second statement but forget the braces, and it still parses but not the way you expect.)

src/Amalgam/entity/EntityExternalInterface.cpp
```diff
@@ -29,20 +29,71 @@ void EntityExternalInterface::LoadEntityStatus::SetStatus(bool loaded_in, std::s
...
+		// We're trying to load into an embedded entity.  First find the existing root
+		auto bundle = FindEntityBundle(handle);
+		if(bundle == nullptr || bundle->entity == nullptr)
+		{
+			status.SetStatus(false, "root entity does not exist");
+			return status;
+		}
+		// Then look up the entity path within the root
+		EvaluableNode *id_path = AssembleEntityPath(bundle->entity->evaluableNodeManager, entity_path);
+		EntityReadReference always_null;
```
howsohazard: I don't think this will be always_null, right?  Because it'll be set to whatever the method below contains.  Consider rename?

src/Amalgam/entity/EntityExternalInterface.cpp
```diff
@@ -68,11 +119,36 @@ EntityExternalInterface::LoadEntityStatus EntityExternalInterface::LoadEntity(st
...
+		std::unique_ptr<std::ostream> log_file = std::make_unique<std::ofstream>(write_log_filename, std::ios::binary);
+		EntityWriteListener *write_log = new EntityWriteListener(entity, std::move(log_file), false, false, false);
 		wl.push_back(write_log);
 	}
 
-	AddEntityBundle(handle, new EntityListenerBundle(entity, wl, pl));
+	if(container != nullptr)
+	{
+		// assigning to a child entity
+		// Are we replacing an existing entity?
```
howsohazard: Shouldn't it load it as an unnamed entity instead of replacing it?

src/Amalgam/entity/EntityExternalInterface.cpp
```diff
@@ -198,6 +295,14 @@ void EntityExternalInterface::StoreEntity(std::string &handle, std::string &path
 	asset_params->UpdateResources();
 
 	asset_manager.StoreEntityToResource(entity, asset_params, true, persistent);
+	if(std::holds_alternative<std::pair<void **, size_t *>>(source))
+	{
+		auto [datap, lenp] = std::get<std::pair<void **, size_t *>>(source);
+		std::string &result = asset_params->resourceContents;
+		*datap = new char[result.size() + 1];
+		memcpy(*datap, result.c_str(), result.size() + 1);
```
howsohazard: I think we use std::memcpy elsewhere in the code?  Double-check.

src/Amalgam/entity/EntityWriteListener.h
```diff
@@ -18,14 +18,14 @@ class EntityWriteListener
 	//if retain_writes is true, then the listener will store the writes, and GetWrites() will return the list of all writes accumulated
 	//if _pretty is true, then the listener will pretty print to filename
 	//if sort_keys is true, then the listener will print with keys sorted for assocs
-	//if filename is not empty, then it will attempt to open the file and log all writes to that file, and then flush the file stream
-	EntityWriteListener(Entity *listening_entity, bool retain_writes = false,
-		bool _pretty = false, bool sort_keys = false, const std::string &filename = std::string());
+	//if transaction_file is not empty, then it will log all writes to that file, and then flush the file stream
```
howsohazard: Note the transfer of ownership in the comment

src/Amalgam/entity/EntityExternalInterface.cpp
dmaze: If I've read `EvaluableNodeTreeFunctions.h` correctly, this will always be null (and I'm glad the name I chose for this variable clearly conveys its intent :)

As I understand `TraverseToEntityReferenceAndContainerViaEvaluableNodeID()`:

* If you pass a null `dest_sid_ref`, then it returns `nullptr, nullptr` if the path doesn't exist; or if the path does exist up to at least the second-to-last entry, it returns `container, entity` where the `entity` could be null.
* If you pass a non-null `dest_sid_ref`, then the "container" entry in the pair is in fact always null, and the second "entity" is either the target entity (and `*dest_sid_ref` is unchanged) or its container (if the path exists only up to the second-to-last entry, and `*dest_sid_ref` is the string ID of the last entry) or null (if the path really doesn't exist).

We're passing in a string ID since we're going to want to eventually inject this as the child, and it really does look like this value will always be null.

src/Amalgam/entity/EntityExternalInterface.cpp
dmaze: "Return an error" would be a reasonable choice.  I don't like the idea of loading the new entity under some different name or path, since the calling code will just lose it.

src/Amalgam/entity/EntityExternalInterface.cpp
dmaze: We probably do because C++, will update this.

src/Amalgam/entity/EntityExternalInterface.cpp
howsohazard: Don't have a way to get the new trainee ID back?  E.g., for temporary and/or unnamed trainees that don't conflict with existing values.  (this is how create_entity and the like work)

src/Amalgam/entity/EntityExternalInterface.cpp
```diff
@@ -29,20 +29,71 @@ void EntityExternalInterface::LoadEntityStatus::SetStatus(bool loaded_in, std::s
...
+		auto bundle = FindEntityBundle(handle);
+		if(bundle == nullptr || bundle->entity == nullptr)
+		{
+			status.SetStatus(false, "root entity does not exist");
+			return status;
+		}
+		// Then look up the entity path within the root
+		EvaluableNode *id_path = AssembleEntityPath(bundle->entity->evaluableNodeManager, entity_path);
+		EntityReadReference always_null;
+		std::tie(always_null, container) = TraverseToEntityReferenceAndContainerViaEvaluableNodeIDPath<EntityReadReference>(bundle->entity, id_path, &id_sid);
```
howsohazard: The template should be EntityWriteReference since you're modifying it.

dmaze: Fixed, thanks.

src/Amalgam/entity/EntityExternalInterface.cpp
dmaze: Looking through other existing code, if you `LoadEntity("handle", "path1.caml", ...)` and then `LoadEntity("handle", "path2.caml", ...)`, without doing anything entity-path related, then the second `LoadEntity()` call will replace the first loaded entity with the second, and the first gets lost.  The behavior in this PR with an entity path – overwrite the sub-entity at that path if it exists – is consistent with that.  There doesn't seem to be anything at this layer around "unnamed trainees".

If there's a specific other behavior you want, it's possible to change it, but I might want to make `LoadEntity()` behave the same for the root entity and for a nested entity.

src/Amalgam/entity/EntityExternalInterface.cpp
dmaze: I've also run this at least once in a debugger and `always_null` is in fact `nullptr`.

src/Amalgam/AmalgamTrace.cpp
```diff
@@ -85,8 +90,59 @@ int32_t RunAmalgamTrace(std::istream *in_stream, std::ostream *out_stream, std::
...
+					rand_seed = command_tokens[7];
+				else
+					rand_seed = random_stream.CreateOtherStreamStateViaString("trace");
+
+				if(command_tokens.size() > 8)
+					entity_path = StringManipulation::Split(command_tokens[8]);
+				else
+					entity_path.clear();
+
+				raw_data = StringManipulation::Base64ToBinaryString(base64_data);
```
howsohazard: Recommend making raw_data a local variable so it is free'd from the heap, whereas the others are short and more frequent, therefore ok to be reused.

howsohazard: the raw_data.clear below will only set the size of the vector back to zero.  You need to call shrink_to_fit in order to release the memory, but at that point better to just make it local.  Same for printing out the data.

src/Amalgam/entity/EntityExternalInterface.cpp
dmaze: We talked this morning and there was consensus that the `LoadEntity()` C API should mirror the `load_entity` Amalgam opcode.  That means, if you pass in an entity path that already exists, the specified entity gets loaded as a child of that entity with a random name.  This in turn also means that `LoadEntity()` needs a path to return the true entity path.

dmaze: This is done.  This _only_ affects the load path (if you're storing a sub-entity, it either exists or it doesn't) and it already had a structured return type, so I added the "correct" entity path there.

src/Amalgam/AmalgamTrace.cpp
dmaze: Done.  I made the same change to `base64_data` for similar reasons.  (And I considered it for `entity_path` but that will be a short vector of short strings.)

src/Amalgam/entity/EntityExternalInterface.h
```diff
@@ -34,12 +35,32 @@ class EntityExternalInterface
...
+	//load/store destinations:
+	struct LoadFromFile { const std::string &path; };
+	struct LoadFromMemory { const std::string &&data; };
+	using LoadSource = std::variant<LoadFromFile, LoadFromMemory>;
+
+	struct StoreToFile { const std::string &path; };
+	struct StoreToMemory { std::string &data; };
+	using StoreSource = std::variant<StoreToFile, StoreToMemory>;
+
+	// Load an entity from some source into a handle.
```
howsohazard, line 50: Some of the comment consists of verbs like "resolves" other part is "load"

src/Amalgam/AmalgamAPI.cpp
```diff
@@ -74,10 +74,22 @@ extern "C"
 
 	LoadEntityStatus ConvertLoadStatusToCStatus(EntityExternalInterface::LoadEntityStatus &status)
 	{
+		size_t entity_path_len = status.entity_path.size();
+		char **entity_path = nullptr;
+		if (entity_path_len > 0)
```
howsohazard: extra space after if

src/Amalgam/Amalgam.h
```diff
@@ -22,6 +22,11 @@ extern "C"
 		bool loaded;
 		char *message;
 		char *version;
+		// If an `entity_path` parameter was passed to LoadEntity, the path where the entity was
```
howsohazard: Are the ` `'s used in any comment comment system?  We don't have any comment system or style in C++ currently that uses them, so it's introducing a new style.

src/Amalgam/AmalgamAPI.cpp
```diff
@@ -74,10 +74,22 @@ extern "C"
...
+		{
+			char *entity_path_alloc = new char[sizeof(char *) * entity_path_len];
+			entity_path = reinterpret_cast<char **>(entity_path_alloc);
```
howsohazard, line 85: Worth calling out the reinterpret cast to be able to use the DeleteString() call

TRACE-FILES.md
```diff
@@ -0,0 +1,138 @@
...
+
+## Syntax
+
+Blank lines and lines that begin with a comment marker, `#` followed by at least one space, are ignored.  Lines with unknown commands or syntactic errors are ignored, and do not stop execution.
+
+Command lines contain a command name followed by some number of space-separated arguments.  An argument may appear in double quotes; if it does, the argument continues through the next matching double-quote.  If a double-quote is needed in an argument value, it may be preceded by a backslash.  No other escaping is possible or allowed, and backslashes that are not followed by double quotes are literally read as backslashes.  If ASCII double-quote bytes 0x22 are escaped by preceding them with a backslash byte 0x5C then an item can otherwise contain arbitrary binary data.
+
+Some commands end with a JSON object.  This is passed directly as a string, and is not quoted or escaped in any way.  Other commands can take a JSON object as a parameter, and this must be quoted and escaped as described above.  Compare:
```
howsohazard, line 13: Should we rectify this?

howsohazard: (not for this PR but in general, and how to best do so?)

src/Amalgam/entity/EntityExternalInterface.h
dmaze: Yes, that's intentional.  In the second paragraph

> If entity_path is non-empty, ... handle must exist, as must all entries in entity_path except for the last one.

and then it talks further about what the path actually resolves to, and where the entity is loaded.  Re-reading the paragraph, I'm pretty sure I consistently use "entity_path resolves" and "load the entity".

src/Amalgam/Amalgam.h
dmaze: Very strong habit from writing Python code with Markdown syntax.  It looks like the style elsewhere here is to use no punctuation or markup at all.  I'll update all of my new comments accordingly.

TRACE-FILES.md
dmaze: Mechanically it wouldn't be a hard change, but it would be a breaking change in the trace-file format.  Since JSON objects have double-quoted strings as object keys, this would mean needing to backslash-escape them everywhere they appear.  This would be a mechanical change in `amalgam-lang-py`, but if you're hand-writing `EXECUTE_ENTITY_JSON` trace lines it'd be a hassle.

Relevant to this PR, I did choose to introduce a bespoke format for the `entity_path` parameters to `LOAD/STORE_ENTITY`.  But maybe it'd be more consistent to pass in a JSON list, load it to an evaluable node, and then chuck it into the underlying machinery.

dmaze: I believe we concluded in a separate chat to not make a change at this time.

howsohazard (approved):

PR #479, opened 2025-12-26
24890: Refactors many of the code mixing opcodes, adding ability to reduce compute for simpler data structures, MAJOR

howsohazard:


src/Amalgam/amlg_code/persist_module_test.amlg
```diff
@@ -1,4 +1,5 @@
-(parallel
+(apply
+	"parallel"
```
howsoRes: why is code with the no longer supported opcode being checked in?

howsoRes (approved):

PR #503, opened 2026-03-08
25156: Adds call_on_entity opcode and changes command line to implement REPL, MINOR

howsohazard:


docs/language.js
```diff
@@ -1741,6 +1741,16 @@ var data = [
...
+		"example" : "(create_entities \"TestContainerExec\"\n  (lambda {d (print \"hello \" x)\n  }) \n)\n\n(print (call_entity \"TestContainerExec\" \"d\" (assoc x \"goodbye\")))"
```
howsoRes: how's this example different from call_entity? can you provide a call_on_entity specific example

src/Amalgam/interpreter/InterpreterOpcodesEntityAccess.cpp
```diff
@@ -466,11 +459,27 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_CALL_ENTITY_and_CALL_ENTIT
...
 
+	auto call_type = en->GetType();
+	StringRef entity_label_sid;
+	EvaluableNodeReference function = EvaluableNodeReference::Null();
+	if(ocn.size() > 1)
+	{
+		if(call_type == ENT_CALL_ON_ENTITY)
+		{
+			function = InterpretNodeForImmediateUse(ocn[1]);
+			node_stack.PushEvaluableNode(function);
```
howsoRes, line 470: does it make sense to early out here if function interpreted to something null or invalid?

howsoRes (approved):

src/Amalgam/interpreter/InterpreterOpcodesEntityAccess.cpp
howsohazard: The only thing that could be checked is if the function were null, anything else requires running the function.  However, even the null check would add a bit of logic to cleanup and should be quite rare, especially because the containing entity is the one sending the code in, and in any evolutionary programming setup, the containing entity is more likely to not be evolved.

PR #516, opened 2026-04-06
25259: Fixes possible assertion failure when using the r command when debugging, removes unnecessary newlines when debugging as well

howsohazard:


fulpm (approved):

src/Amalgam/interpreter/InterpreterDebugger.cpp
fulpm: Do we need the extra new line here too?

PR #570, opened 2026-06-02
25573: Improves documentation and spell checking process

howsohazard:


howsoRes (approved):

domg123-hermes:
One thing I’d double-check in the new `unparse` / `print` descriptions: the `@` emission part looks conditional on `include_attributes`, but the new wording reads unconditional.

Specifically, `InterpretNode_ENT_UNPARSE` defaults `include_attributes` to false and passes it through to `Parser::Unparse(tree, pretty, include_attributes, ...)`. In `Parser::Unparse`, the `@` marker for `preevaluationNeeded` is only appended inside the `if(upd.emitAttributes)` block. `print` also calls `Parser::Unparse(cur, true, false, true)`, so it looks like it is safe for cyclic/graph structures but does not emit `@` markers for parse-reconstructing those relationships by default.

So perhaps these should be softened to something like: cyclic/graph unparsing is safe, and parse-reconstructing `@` output is available when attributes are included; `print` is safe but may not produce reconstructing `@` output.

PR #584, opened 2026-06-22
24886: Add `query_entity_clusters` opcode, MINOR

dmaze:
Add a `query_entity_clusters` opcode that performs HDBSCAN clustering over the entities, returning the cluster IDs.  Note that the clustering uses the standard surprisal-based distance metric, which produces different results from the same clustering algorithm using Euclidean distances.


src/Amalgam/entity/EntityQueriesClustering.cpp
```diff
@@ -0,0 +1,96 @@
+//project headers:
+#include "EntityQueriesDensityFunctions.h"
+#include "HDBSCANClustering.h"
```
howsohazard: File should be named EntityQueriesClustering.* to fit convention

src/Amalgam/entity/EntityQueriesClustering.cpp
```diff
@@ -0,0 +1,96 @@
...
+		else //treat as infinite distance if no neighbors (not included in query)
+			core_distances[entity] = std::numeric_limits<double>::infinity();
+	}
+#ifdef MULTITHREAD_SUPPORT
+		, runConcurrently
+#endif
+	);
+
+	//dense 0..m-1 indexing over the entities we are clustering
+	std::vector<size_t> entity_to_dense(num_entity_indices, std::numeric_limits<size_t>::max());
```
howsohazard: Do these lookup structures improve readability or performance?  Most of the time the data points being clustered are dense (the vast majority of the records).

src/Amalgam/entity/EntityQueriesClustering.cpp
```diff
@@ -0,0 +1,96 @@
...
+		dense_to_entity.push_back(entity);
+	}
+	size_t m = dense_to_entity.size();
+
+	//per-point weights (summed mass uses these)
+	std::vector<double> point_weights(m, 1.0);
+	for(size_t d = 0; d < m; ++d)
+	{
+		double w = 1.0;
+		distanceTransform->getEntityWeightFunction(dense_to_entity[d], w);
```
howsohazard: The getEntityWeightFunction method is a handful of branches and two lookups.  Is it more performant to precache these?  If so, it should be clearly stated why.  If not, don't precache.

src/Amalgam/interpreter/OpcodesEntityQueryEngine.cpp
```diff
@@ -2111,6 +2111,49 @@ static OpcodeInitializer _ENT_QUERY_ENTITY_CUMULATIVE_NEAREST_ENTITY_WEIGHTS(ENT
...
...s local density (its core distance); it should be smaller than the expected cluster size.  min_cluster_weight is the smallest accumulated entity weight that will be considered a cluster, and can be 0.  The ids are nonnegative integers, with the id of zero denoting entities that are independent and isolated from all clusters (noise).  feature_labels specifies the names of the features to consider during computation.  See Distance and Surprisal Calculations for details on the other parameters and how distance is computed.  If `output_sorted_list` is not specified or is false, then it will return an assoc of entity string id as the key with the cluster id as the value; if `output_sorted_list` is true, then it will return a list of lists, where the first list is the entity ids and the second list contains the corresponding cluster ids, both in order of ascending cluster id. If `output_sorted_list` is a string, then it will additionally return a list where the values correspond to the values of the labels for each respective entity.  If `output_sorted_list` is a list of strings, then it will additionally return a list of values for each of the label values for each respective entity.)";
```
howsohazard: backticks should be consistent around parameter names

build/cmake/create_tests.cmake
```diff
@@ -95,6 +95,21 @@ foreach(TEST_TARGET ${ALL_SHAREDLIB_TARGETS})
 
 endforeach()
 
+# Standalone pure-algorithm unit tests (no Amalgam library link).
+# Skipped under WASM: the global emscripten linker flags export the C API
+# symbols (LoadEntity, etc.), which this library-free executable cannot supply.
+if(NOT IS_WASM)
+    add_executable(hdbscan-unit-test "test/unit_test/hdbscan_test.cpp")
```
howsohazard: Why add another unit test executable?  Would be better to just combine with the current unit test.

src/Amalgam/entity/HDBSCANClustering.h
```diff
@@ -0,0 +1,474 @@
...
+	//One internal node of the single-linkage dendrogram (one per accepted MST edge).
+	struct SingleLinkageNode
+	{
+		size_t left;	//child node id (point id < m, or internal node id >= m)
+		size_t right;	//child node id
+		double weight;	//mutual-reachability distance at which the children merged
+		double mass;	//summed point weight of this subtree
+	};
+
+	//A child leaving cluster `parent` at density level `lambda`.
```
howsohazard: Current Amalgam style is to not use back ticks around variables in comments

src/Amalgam/entity/HDBSCANClustering.h
```diff
@@ -0,0 +1,474 @@
...
+	{
+		size_t u;
+		size_t v;
+		double weight;
+	};
+
+	//One internal node of the single-linkage dendrogram (one per accepted MST edge).
+	struct SingleLinkageNode
+	{
+		size_t left;	//child node id (point id < m, or internal node id >= m)
```
howsohazard: What is "m"?

src/Amalgam/entity/HDBSCANClustering.h
```diff
@@ -0,0 +1,474 @@
...
+		size_t right;	//child node id
+		double weight;	//mutual-reachability distance at which the children merged
+		double mass;	//summed point weight of this subtree
+	};
+
+	//A child leaving cluster `parent` at density level `lambda`.
+	struct CondensedEdge
+	{
+		size_t parent;		//cluster label (>= m)
```
howsohazard: What is "m" here (also see prior comment)?  Should be clearly defined somewhere.

src/Amalgam/entity/HDBSCANClustering.h
```diff
@@ -0,0 +1,474 @@
...
+	{
+		std::sort(edges.begin(), edges.end(),
+			[](const Edge &a, const Edge &b) { return a.weight < b.weight; });
+
+		std::vector<size_t> parent(m);
+		std::vector<size_t> rank(m, 0);
+		for(size_t i = 0; i < m; ++i)
+			parent[i] = i;
+
+		std::function<size_t(size_t)> find = [&](size_t x) -> size_t
```
howsohazard: The inclusion of std::function adds overhead.  Use auto here instead.

src/Amalgam/entity/HDBSCANClustering.h
```diff
@@ -0,0 +1,474 @@
...
+	//Builds the minimum spanning tree (forest, if disconnected) via Kruskal's
+	//algorithm.  Returns accepted edges in ascending weight order.
+	inline std::vector<Edge> BuildMST(size_t m, std::vector<Edge> edges)
+	{
+		std::sort(edges.begin(), edges.end(),
+			[](const Edge &a, const Edge &b) { return a.weight < b.weight; });
+
+		std::vector<size_t> parent(m);
+		std::vector<size_t> rank(m, 0);
+		for(size_t i = 0; i < m; ++i)
```
howsohazard: Current amalgam style is that if the iterator is simple, use a suffix ++, e.g., i++.  This applies elsewhere too.

src/Amalgam/entity/EntityQueriesClustering.h
```diff
@@ -0,0 +1,474 @@
...
+				std::swap(ra, rb);
+			parent[rb] = ra;
+			if(rank[ra] == rank[rb])
+				++rank[ra];
+		}
+		return mst;
+	}
+
+	//Summed point weight of a node (point if < m, else internal subtree).
+	inline double NodeMass(size_t node, size_t m,
```
howsohazard, line 108: m should be clearly defined (see other comments)

src/Amalgam/entity/HDBSCANClustering.h
```diff
@@ -0,0 +1,474 @@
...
+	inline std::vector<SingleLinkageNode> BuildSingleLinkageTree(size_t m,
+		std::vector<Edge> mst, const std::vector<double> &point_weights)
+	{
+		std::vector<SingleLinkageNode> nodes;
+		nodes.reserve(mst.size());
+
+		std::vector<size_t> parent(m);
+		for(size_t i = 0; i < m; ++i)
+			parent[i] = i;
+		std::function<size_t(size_t)> find = [&](size_t x) -> size_t
```
howsohazard: For performance, use auto not std::function

src/Amalgam/entity/HDBSCANClustering.h
```diff
@@ -0,0 +1,474 @@
...
+			comp_node[ra] = new_id;
+		}
+		return nodes;
+	}
+
+	//Appends every leaf point id under `node` to `out`.
+	inline void CollectLeaves(size_t node, size_t m,
+		const std::vector<SingleLinkageNode> &nodes, std::vector<size_t> &out)
+	{
+		std::vector<size_t> stack;
```
howsohazard: Seems inefficient to allocate a new stack every call instead of reusing a buffer.

src/Amalgam/entity/HDBSCANClustering.h
```diff
@@ -0,0 +1,474 @@
...
+		for(const SingleLinkageNode &sn : nodes)
+		{
+			is_child[sn.left] = true;
+			is_child[sn.right] = true;
+		}
+
+		size_t next_label = m;	//condensed cluster labels live above point ids
+		FastHashMap<size_t, size_t> node_label;	//dendrogram node id -> cluster label
+
+		std::vector<size_t> work;	//internal node ids still to process
```
howsohazard: "work" is not a good variable name

src/Amalgam/entity/HDBSCANClustering.h
```diff
@@ -0,0 +1,474 @@
...
+		size_t total_nodes = m + nodes.size();
+		std::vector<bool> is_child(total_nodes, false);
+		for(const SingleLinkageNode &sn : nodes)
+		{
+			is_child[sn.left] = true;
+			is_child[sn.right] = true;
+		}
+
+		size_t next_label = m;	//condensed cluster labels live above point ids
+		FastHashMap<size_t, size_t> node_label;	//dendrogram node id -> cluster label
```
howsohazard: don't just use the term "label" as it has a well-defined meaning elsewhere in amalgam.  call it cluster_id or something more specific

src/Amalgam/entity/HDBSCANClustering.h
```diff
@@ -0,0 +1,474 @@
...
+				//threshold; smaller components (e.g. scattered outliers) stay noise
+				if(NodeMass(id, m, point_weights, nodes) >= min_cluster_weight)
+				{
+					node_label[id] = next_label++;
+					work.push_back(id);
+				}
+			}
+		}
+
+		while(!work.empty())
```
howsohazard: This while block could use a few more comments explaining what is going on

src/Amalgam/entity/HDBSCANClustering.h
```diff
@@ -0,0 +1,474 @@
...
+					}
+					else	//a heavy continuing leaf stays in `label` to the bottom
+						result.push_back(CondensedEdge{label, child, lambda_max, point_weights[child]});
+				}
+			}
+		}
+		return result;
+	}
+
+	//Birth lambda for every cluster label.  A cluster is born at the lambda of
```
howsohazard: This code is slightly hard to follow for someone not closely familiar with the algorithm.  It may need a little better description.

src/Amalgam/entity/HDBSCANClustering.h
```diff
@@ -0,0 +1,474 @@
...
+	//`stability` and `birth` are taken by value: both are moved-in maps consumed and
+	//freed when this returns, since selection is their last reader.  `birth` is the
+	//ClusterBirthLambdas map, shared with ComputeStabilities so it is computed once.
+	inline FastHashSet<size_t> SelectClusters(size_t m,
+		const std::vector<CondensedEdge> &condensed,
+		FastHashMap<size_t, double> stability,
+		FastHashMap<size_t, double> birth)
+	{
+		//cluster -> parent cluster, and cluster -> child clusters
+		FastHashMap<size_t, size_t> cluster_parent;
```
howsohazard: Are hash maps the fastest data structures for cluster_parent?  Or would parallel vectors or vectors of cluster id with parent be more efficient?

src/Amalgam/entity/EntityQueriesClustering.cpp
dmaze: Renamed it.  The header file includes the entire class and several related capabilities and so it doesn't change.

src/Amalgam/interpreter/OpcodesEntityQueryEngine.cpp
dmaze: It looks like other docs don't use backticks at all, so removed them.

src/Amalgam/entity/HDBSCANClustering.h
dmaze: Fixed.

src/Amalgam/entity/HDBSCANClustering.h
```diff
@@ -0,0 +1,474 @@
...
+		double mass;	//summed point weight of this subtree
+	};
+
+	//A child leaving cluster `parent` at density level `lambda`.
+	struct CondensedEdge
+	{
+		size_t parent;		//cluster label (>= m)
+		size_t child;		//point id (< m) or sub-cluster label (>= m)
+		double lambda;		//1 / weight at which the child separates from parent
+		double child_mass;	//summed weight of the child
```
howsohazard: Amalgam convention is to use lowerStartingCamelCase for all attributes and reserves snake_case for the stack.  This should be childMass.

src/Amalgam/entity/HDBSCANClustering.h
howsohazard: For naming convention rather than left and right, it should be something like leftIndex, leftId, or whatever is more descriptive to how to look it up (and should conform to Amalgam style for attributes using lowerStartingCamelCase).

src/Amalgam/entity/HDBSCANClustering.h
howsohazard: the parent and child variables should be more descriptive as to how to look them up and should conform to Amalgam style for attributes using lowerStartingCamelCase, e.g., parentId, parentIndex

build/cmake/create_tests.cmake
dmaze: This got folded into the existing C++ unit-test rig.

src/Amalgam/entity/EntityQueriesClustering.cpp
dmaze: It doesn't seem like it.  It's being argued to me that this is required for correctness, but the arguments for that are that the vector size is used as a sentinel value, and also

> Without it you'd size everything by the max index and the gap indices would become phantom isolated "points" that change the result (extra noise, inflated m, skewed mass/threshold logic).

There's been some cleanup in this space (the reverse `dense_to_entity` mapping was removed) but I will revisit this concept.

src/Amalgam/entity/EntityQueriesClustering.cpp
dmaze: This is still computed once up front and passed in as a parameter to the separated HDBSCAN code.  Arguments for doing this are that the weights are used multiple times, but more notably that if we precompute the per-entity weights then we do not need to keep the entity vector.

This is a second place I need to reason through this a little bit more.

src/Amalgam/entity/HDBSCANClustering.h
dmaze: A comment got added above this stating explicitly "m is the number of input points being clustered."

src/Amalgam/entity/HDBSCANClustering.h
dmaze: Fixed.

src/Amalgam/entity/HDBSCANClustering.h
dmaze: Fixed everywhere.

src/Amalgam/entity/HDBSCANClustering.h
dmaze: Changed this so the `stack` vector is passed as a parameter and can be reused from the caller.

src/Amalgam/entity/HDBSCANClustering.h
dmaze: Now named `pending_nodes`.

src/Amalgam/entity/HDBSCANClustering.h
dmaze: Now named `node_cluster_id`.

src/Amalgam/entity/HDBSCANClustering.h
dmaze: Gained an extended comment block, starting with "Walk each pending cluster node down the dendrogram."

src/Amalgam/entity/HDBSCANClustering.h
dmaze: This got expanded out a little bit.  I think the description is fine (as distinctly Not An Expert).  From one random Internet-found writeup of [How HDBSCAN Works](https://hdbscan.readthedocs.io/en/latest/how_hdbscan_works.html) the λ value is pretty prominent, and I don't think you need _that_ deep algorithmic knowledge for this to make sense.

src/Amalgam/entity/HDBSCANClustering.h
dmaze: These almost universally got changed to `std::vector<>`.  (...and now the implementer wants to keep these vectors very dense, but that's its own issue.)

src/Amalgam/entity/HDBSCANClustering.h
dmaze: Fixed.

src/Amalgam/entity/HDBSCANClustering.h
dmaze: `leftChildId` and `rightChildId` now.

src/Amalgam/entity/HDBSCANClustering.h
dmaze: `parentId` and `childId` now.

src/Amalgam/entity/EntityQueriesClustering.cpp
dmaze: I've had this mapping removed.  The tradeoff is that some of the internal structures will always have size equal to the number of entities, even if you're not running clustering on all of them.

src/Amalgam/entity/EntityQueriesClustering.cpp
dmaze: I don't think there's an advantage to precaching these, and we were able to restructure the pure clustering implementation to take a function mapping entity index to weight instead of passing in a vector.

howsohazard (approved):

PR #585, opened 2026-06-23
25660: Improvements to flexibility of mutate opcodes, changes and expands mutation typse, MAJOR

howsohazard:


domg123:
Reviewed this branch (built locally as `80.0.6-alpha`) from the GP side — this is exactly the set of mutation levers we'd been after, and the `.null`-key-preserves-default design for `immediate_number_weights` / `immediate_string_weights` is clean. I confirmed the categorical numeric draw works (e.g. `immediate_number_weights {101 1 201 1 301 1}` yields those values throughout). Two things I think are bugs, plus one heads-up.

### 1. Missing `break;` — `change_type` falls through into `insert`
`EvaluableNodeTreeManipulation.cpp` ~L1274–1282: the `case ENBISI_change_type:` block closes without a `break;`, so every `change_type` mutation falls through into `ENBISI_insert` and additionally wraps the node in a freshly-allocated parent. Pre-PR, `change_type` ended with `break;`.

Repro on this branch:
```
(mutate 5 1.0 .null (associate "change_type" 1))
```
Over 300 trials, 9 results have `total_size > 1` — a `change_type` on a leaf can't give it children, so it should be 0/300. It only *surfaces* when the inserted wrapper happens to be a container type; when the wrapper is an immediate the appended child is hidden, so the real fall-through rate is ~100%. Visible examples from mutating the leaf `5`: `(current_value s8b)`, `[SbnqkS4X]`.

Effect: `change_type` (the highest-weight op at 0.15) is silently doubled with an insert, biasing the whole mutation distribution toward growth.

### 2. `replace_element_with_copy` on an assoc looks inverted / double-applies
`EvaluableNodeTreeManipulation.cpp` ~L1343–1377 (mapped-array branch). `destination_index` is drawn from `[0, num_children]`, with the `+1` slot meant to mean "add a new key". But:

- When it matches an existing entry (`destination_index < 1`), it replaces that value and then runs `destination_index--`. Since `destination_index` is a `size_t` already at 0, this underflows to `SIZE_MAX`, so the trailing `if(destination_index > 0)` is true and it *also* inserts a new key — replace **and** add in one op.
- When `destination_index == num_children` (the intended "new key" case), the loop drains it to exactly 0, so `if(destination_index > 0)` is false and no new key is created — the opposite of the intent.

The in-loop `destination_index--` likely wants to be `destination_index = 0` (or a found-flag). Repro: mutating `{a 1 b 2 c 3}` with `replace_element_with_copy`-only yields single-step results where the copied value lands in an overwritten existing key **and** a freshly added key at once, e.g. `{N81pbCC 0.69… a 0.34… b 0.69… c 0.69…}`.

### Heads-up (not a bug)
Mutation is now applied post-order — `MutateTree` moved the `Rand() < mutation_rate` / `MutateNode` call from *before* recursing into children to *after*. That changes the result of any seeded mutation vs the previous engine, so cross-version golden/determinism tests will shift. Looks intentional, just flagging the compat surface.

howsohazard:
The two bugs found in my PR mentioned in the previous comment have been fixed.

domg123:
Re-tested after `81f0fb0` (rebuilt locally, `80.0.6-alpha`). **Both original bugs are fixed** — `change_type` no longer falls through (leaf-growth back to 0/300), and `replace_element_with_copy` on an assoc no longer double-applies (replace-over-existing is clean). Two more things surfaced while verifying, both small:

### 3. Stray comma in the interned op name (`Opcodes.cpp:441`)
```cpp
EmplaceStaticString(ENBISI_replace_element_with_copy, "replace_element_with_copy,");
```
The string literal has a trailing comma, so the op registers as `replace_element_with_copy,` rather than `replace_element_with_copy`. Two effects:
- Passing the **documented** name `operation_type {replace_element_with_copy 1}` matches nothing; if it's the only key the weights map ends up empty and mutation silently falls back to the full default op-mix (this is what tripped up my first repro).
- `(get_mutation_defaults "mutation_types")` returns the key with the comma — visible in the PR's own doc example: `replace_element_with_copy, 0.0999`.

Fix is just dropping the comma in the literal.

### 4. `replace_element_with_copy` on an assoc never adds a new key
In the mapped-array branch, after the destination loop:
```cpp
//if didn't find destination_index, need to create a new key
if(cur_index < destination_index)
```
`cur_index` ends equal to `destination_index` in **both** outcomes — when the destination is found (breaks at `cur_index == destination_index`) and when it isn't (the `destination_index == num_children` "new key" case, where the loop runs to completion leaving `cur_index == num_children == destination_index`). So `cur_index < destination_index` is never true and the new-key branch is dead. Net: assoc `replace` only ever overwrites an existing value, never appends a new key — whereas the ordered/list branch's `if(destination_index >= num_children)` append works fine.

Verified by isolating the op (via its actual comma'd name) on `{a 1 b 2 c 3}`, 400 trials: keys=3 → 400/400, keys=4 → 0/400; the same probe on `[1 2 3]` grows to 4+ elements ~16/400.

Suggested fix: mirror the ordered branch and guard the new-key insert with `if(destination_index >= num_children)` instead.

(Both are minor and don't affect our own usage yet — we pass `operation_type` as null — just flagging for completeness.)

howsohazard:
Pushed fixes to those two bugs.

domg123:
Re-tested after `f02d541` (rebuilt locally). **#1, #2, and #3 are all verified fixed** — `change_type` doesn't grow leaves (0/300), no double-apply on assocs, and the comma is gone so `replace_element_with_copy` is now addressable by its documented name (`get_mutation_defaults` is clean too).

**#4 is half-fixed — the dead-code path now fires, but it copies a null.** The assoc new-key branch now runs (4-key results appear ~111/400), but `source_node` is only located inside the `else` (replace) branch, so in the new-key branch it's still `nullptr`:

```cpp
EvaluableNode *source_node = nullptr;
...
if(destination_index >= num_children)
{
    std::string new_key = GenerateRandomStringGivenStringSet(...);
    n->SetMappedChildNode(new_key, mp.enm->DeepAllocCopy(source_node));  // source_node is nullptr here
}
else //replace an existing element
{
    // ... source-finding loop only runs in this branch ...
}
```

So `replace_element_with_copy` on an assoc adds `{new_key: .null}` instead of a copy of an existing element — it no longer *copies* anything in that case. Samples (documented name, only 4-key results shown):
```
{"3k8n" .null a -2.63… b 0 c 3.82…}
{a 1 b 2 c 2 fd6F .null}
{_SEhVSewEL .null a 0 b -0.27… c 1}
```
Every new key is `.null`; 111/111 of the 4-key results had a null new value (0 real copies).

Fix: hoist the source-finding loop above the `if`, so `source_node` is populated for both the new-key and replace paths (the destination loop can stay in the `else`). (No `DeepAllocCopy(nullptr)` crash on this build, but it's also a null-deref waiting to bite if `DeepAllocCopy` ever stops null-guarding.)

domg123 (approved):
Re-tested after `6835a28` — **all five issues from this review are resolved**, verified on a local build:

- **#1** `change_type` no longer falls through (leaf-growth 0/300).
- **#2** `replace_element_with_copy` on assocs no longer double-applies.
- **#3** comma dropped — the op is addressable by its documented name and `get_mutation_defaults` is clean.
- **#4** the assoc new-key path now fires (~31% of assoc replaces, matching the expected ~1/num_children).
- **#5** the new key now carries a real copied value — 187 four-key results, 0 containing a null (source-finding loop hoisted above the branch as suggested).

Also ran 2000 default-op mutations over a mixed tree (lists/assocs/opcodes/immediates) with no crash, and confirmed the new `immediate_number_weights` / `immediate_string_weights` levers work as documented.

Nice set of additions — these are exactly the mutation primitives we'd been after on the GP side. LGTM. (The deeper "can native mutate/mix stand in for our own graft" testing is the post-release follow-up we discussed; nothing blocking here.)

howsoRes (approved):

PR #587, opened 2026-06-25
25678: Fixes string leak in mutate operation, fixes doc issues

howsohazard:


src/Amalgam/evaluablenode/EvaluableNode.cpp
```diff
@@ -995,6 +994,10 @@ EvaluableNode *EvaluableNode::EraseMappedChildNode(const StringInternPool::Strin
 	string_intern_pool.DestroyStringReference(sid);
 	EvaluableNode *erased_value = found->second;
 	mcn.erase(found);
+
+	if(!GetIsIdempotent() && mcn.size() == 0)
```
domg123, line 998: Non-blocking question: this recomputes idempotency only when the assoc hits empty (`mcn.size() == 0`). Removing a non-last child that was the sole non-idempotent one wouldn't upgrade the flag here. Since the flag is conservative (a stale `false` is always safe) this is a benign missed-optimization rather than a bug — just confirming the `size() == 0` gate is intentional and not meant to be more general. Also worth noting this now runs for every `EraseMappedChildNode` caller, slightly beyond the mutate-leak scope (fine, arguably more correct).

domg123 (approved):
Reviewed the string-leak fix — correct and minimal. Routing the assoc-key erase through \`EraseMappedChildNode\` properly releases the interned string reference that the raw \`mcn.erase(key)\` was leaking. Checked the other mapped-child erase sites (OpcodesIteration 1199/1207, Entity.cpp:401, the container-move at OpcodesContainerManipulation:2456) — all either destroy the string ref explicitly or hand it off, so MutateNode was the lone offender; the fix is complete. Doc edits correctly scrub the stale \`call_sandboxed\` references. One non-blocking question left inline on the idempotency block.

src/Amalgam/evaluablenode/EvaluableNode.cpp
howsohazard: Yes, that is correct.

PR #589, opened 2026-06-25
00000: Clarify .null idioms help, PATCH

howso-mario:
## Summary
- Updates the source-level built-in idioms help string in `src/Amalgam/interpreter/OpcodesSystemAndRuntime.cpp` to clarify `.null` sentinel guidance.
- Regenerates `docs/idioms.md` from `build/docs/build_doc_files.amlg` so generated docs match the source help string.
- Keeps the change narrow: one paragraph in the idioms help topic.

## What was missing
The previous idioms help paragraph only covered implicit `.null` from an `if` without an else and warned against overloading `.null`. It did not mention `.null` as the immediate no-value sentinel, `get` missing/out-of-bounds behavior, `get_type_string` for runtime checks, or the need for tagged/state wrappers when outcomes must stay distinct.

## Deep understanding summary
From the language source: `.null` is the immediate null/no-value value (`ENT_NULL`; `InterpretNode_ENT_NULL` returns `EvaluableNodeReference::Null()`). `InterpretNode_ENT_IF` returns null when no condition matches and no else exists. `InterpretNode_ENT_GET` returns null when a requested destination/path is absent. `InterpretNode_ENT_GET_TYPE_STRING` defaults null references to `ENT_NULL` and returns the readable type string `"null"`. Live probe with the locally built interpreter confirmed `(/ 0 0)`, `(+ .null 1)`, and `(concat .null)` return `.null`, and `get_type_string` reports `null`. The guidance therefore frames `.null` as a no-value sentinel, not as `0`, `""`, or `[]`, and recommends tagged/state wrappers when `.null` is valid data or result states need to stay distinct.

## Verification
- Built local debug interpreter: `cmake --preset amd64-debug-linux -DCMAKE_BUILD_WITH_INSTALL_RPATH=ON` and `cmake --build --preset amd64-debug-linux --target amalgam-mt-app -j2`.
- Generated docs with the built interpreter: `cd docs && ../out/build/amd64-debug-linux/amalgam-mt ../build/docs/build_doc_files.amlg`.
- Verified built `(help "idioms")` and `docs/idioms.md` contain the new `.null` guidance.
- Ran `./out/build/amd64-debug-linux/amalgam-mt --validate-amalgam` → `All Tests Passed`.
- Independent Claude review after cleanup: no blocking findings.

## Notes
- The local build used temporary `gcc-10`/`g++-10` wrapper symlinks to the available GCC 14 compiler because this container does not have GCC 10 installed; no repo files were changed for that workaround.
- Human review/merge required.

### DIFF
--- a/docs/entity_query_engine.md
+++ b/docs/entity_query_engine.md
@@ -2226,3 +2226,53 @@ Output:
 
 [Amalgam Opcodes](./opcodes.md)
 
+### Opcode: `query_entity_clusters`
+#### Parameters
+`list|number selection_bandwidth  list feature_labels number min_cluster_weight [number p_value] [list|assoc|assoc of assoc weights] [list|assoc distance_types] [list|assoc attributes] [list|assoc deviations] [string weights_selection_feature] [string|number distance_transform] [string entity_weight_label_name] [number random_seed] [string radius_label] [string numerical_precision] [* output_sorted_list]`
+#### Description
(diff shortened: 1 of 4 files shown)


howso-mario:
Closing in favor of #590. This PR included generated docs in the diff (`docs/idioms.md` and related generated markdown), but those docs are produced from source help strings and should not be hand-edited or committed directly. The replacement PR keeps the same `.null` idioms help improvement at the source level only (`src/Amalgam/interpreter/OpcodesSystemAndRuntime.cpp`).

PR #595, opened 2026-06-30
25699: Improves performance, reduces memory, fixes incorrect reporting issue in memory performance counters

howsohazard:


mkoistinen (approved):
Approval we we can build

apbassett (approved):

PR #603, opened 2026-07-06
25732: Fixes three crashes found from evolutionary fuzz testing

howsohazard:


src/Amalgam/interpreter/Interpreter.h
```diff
@@ -120,7 +124,33 @@ class Interpreter
...
+
+				//if something could have stored the target somewhere, then can't be unique
+				if(execution_side_effects)
+				{
+					target_ref.unique = false;
+					target_ref.uniqueUnreferencedTopNode = false;
+				}
+
+				//clear freeability for use in other places
+				target_ref->SetIsFreeable(false);
```
domg123-hermes, line 155: In the concurrent opcodes (`map`/`filter`/`associate`/`list`/`range`), every worker's construction-stack entry holds the same `target` pointer (`&result`), so each worker's `Pop()` writes the shared result reference's node attributes through the non-atomic `SetNeedCycleCheck`/`SetIsFreeable` while workers hold only a shared read lock — concurrent non-atomic read-modify-writes on one `attributes` word.

It looks benign in practice (the writes converge to the same values, and GC's attribute writes are excluded by the write lock), but the `ENT_TARGET`/`ENT_ARGS` paths added in this PR use `SetIsFreeableAtomic` for what appears to be the same shared-node concern. Is the non-atomic setter intentional here? If so, a one-line note on why it's safe would help; otherwise the atomic variant would match.

src/Amalgam/interpreter/Interpreter.cpp
```diff
@@ -715,11 +724,14 @@ void Interpreter::PopulatePerformanceCounters(InterpreterConstraints *interprete
...
-		if(remaining_depth > 0)
+
+		//setting up a new interpreter to populate the constraints requires at least one node
+		//need to take it away, but can't let it go to 0, so if at last one, just end execution
+		if(remaining_depth > 1)
 		{
 			if(interpreter_constraints->ConstrainedOpcodeExecutionDepth())
 				interpreter_constraints->maxOperationDepth = std::min(
-					interpreter_constraints->maxOperationDepth, remaining_depth);
+					interpreter_constraints->maxOperationDepth, remaining_depth - 1);
```
domg123-hermes: The constrained branch now reserves one (`remaining_depth - 1`), but the `else` just below still assigns `remaining_depth` unchanged. If the "take one away for the new interpreter" rationale applies generally, should the unconstrained case subtract one too — or is it deliberately exempt?

src/Amalgam/evaluablenode/EvaluableNodeManagement.h
```diff
@@ -31,13 +32,15 @@ struct ConstructionStackEntry
 	{}
 
 	EvaluableNode *targetOrigin;
-	EvaluableNode *target;
+	EvaluableNodeReference *target;
```
domg123-hermes: On the rename raised in the description: since this now points to the caller's reference rather than the target node itself, a rename does help — the `cs_entry.target` vs `*cs_entry.target` distinction is where this bug class hides. `targetRefPtr` works; `targetLocation` or `targetReference` may read more consistently alongside `targetOrigin`.

domg123-hermes (commented):
Built this branch and ran an internal Amalgam program corpus (~190 symbolic-regression, boolean, and rosetta-style scripts) against it, with no regressions vs `main`.

Separately, it resolves a reliably-reproducing SIGSEGV in a nested capped `call_entity` (sandboxed evaluation of a dangling contained-entity reference). Bisecting places the fix at the constraint-limiter commit, which confirms that path was a genuine memory-safety issue and not only a limiter bypass.

A few non-blocking points inline: a concurrency question on the new `Pop()` attribute writes, a small depth-accounting question, and a note on the `target` field naming.

domg123 (approved):

PR #614, opened 2026-07-17
25793: Improves mix and mutate opcodes

howsohazard:


src/Amalgam/evaluablenode/EvaluableNodeTreeManipulation.cpp
```diff
@@ -483,16 +547,63 @@ EvaluableNode *EvaluableNodeTreeManipulation::MergeTrees(NodesMergeMethod *mm, E
...
+	//try to record, but if already checked, then don't do anything
+	auto [_, inserted] = strings_from_tree_data.checked.insert(tree);
+	if(!inserted)
+		return;
+
+	if(tree->IsAssociativeArray())
+	{
+		for(auto &[cn_id, cn] : tree->GetMappedChildNodesReference())
+		{
+			strings_from_tree_data.keyAndSymbolStrings.push_back(cn_id->string);
```
domg123-hermes: **`mutate_entity` / `mix` segfault on essentially any entity.** `cn_id` is the assoc key, and an entity stores its unlabeled root code under the **null key** (`NOT_A_STRING_ID`), so `cn_id->string` dereferences null.

Repro: `(create_entities "E" (lambda (declare (assoc x 1 y 2) (+ (get_value "x") (get_value "y")))))` then `(mutate_entity "E" 0.3 "M")` faults here before any mutation happens. lldb: `cn_id == nullptr`, fault address `0x1f` = null + the offset of `std::string` inside `StringInternStringData`.

Also worth noting: the `if(tree == nullptr) return;` guard that opened the old `GetStringsFromTree` was dropped in this rewrite.

Guarding the key read fixes it:
```cpp
if(cn_id != string_intern_pool.NOT_A_STRING_ID)
    strings_from_tree_data.keyAndSymbolStrings.push_back(cn_id->string);
```
This is almost certainly the crash you mentioned pushing a fix for — flagging the precise trigger (the null root key) in case it helps confirm coverage.

domg123-hermes (commented):
Automated review of the mix/mutate changes (Dom's agent). I built a debug engine at this head (83c9afd) and reproduced each item below with a small script + lldb; happy to share the repros.

I left the `docs/` diff alone — those look like the auto-rebuild picking up already-merged content from the prior PR, per Chris's note, not new changes to review.

Most of these live in the new `call_entity` repair. Note that #1 (the crash) sits upstream of that repair, so #3 and #4 aren't reachable at runtime until the crash is cleared — which is likely why they slipped through. #1 itself is very probably the crash you mentioned already fixing; I've included the exact trigger in case it's useful.

src/Amalgam/evaluablenode/EvaluableNodeTreeManipulation.h
```diff
@@ -60,37 +60,36 @@ class EvaluableNodeTreeManipulation
...
-				interpreter(nullptr),
-				enm(nullptr),
-				mutation_rate(0),
-				strings(nullptr),
+			WeightedRandValueType &imm_string_weights)
+				: interpreter(interpreter), enm(_enm), entity(_entity), mutation_rate(mutation_rate),
+				keyAndSymbolStrings(key_and_symbol_strings), valueStrings(value_strings),
 				references(EvaluableNode::ReferenceAssocType()),
-				randEvaluableNodeType(nullptr),
+				randEvaluableNodeType(rand_operation),
```
domg123-hermes, line 76: **Dropped assignment in the constructor refactor — `operation_type` weights are now ignored.** Moving to the init-list, `randEvaluableNodeType(rand_operation)` (this line) and `preserveTypeDepth(preserve_type_depth)` were correctly wired to their params, but the line below keeps `randMutationType(&mutationOperationTypeRandomStream)` (the static default). The old body's `this->randMutationType = rand_operation_type;` is gone, so the `rand_operation_type` argument is now unused.

`randMutationType` selects the mutation operation (read in `MutateNode`, ~L1313 / L1338), so a caller-supplied operation-type weighting to `mutate` / `mutate_entity` now has no effect — mutation always uses the default operation distribution.

Fix is on the next line:
```cpp
randMutationType(rand_operation_type),
```

src/Amalgam/evaluablenode/EvaluableNodeTreeManipulation.cpp
```diff
@@ -1630,16 +1684,91 @@ EvaluableNode *EvaluableNodeTreeManipulation::MutateTree(MutationParameters &mp,
...
+					TraverseToExistingEntityReferenceViaEvaluableNodeIDPath<EntityReference<Entity>>(mp.entity, n_ocn[0]);
+
+				//if valid entity, use it
+				if(attempted_entity != nullptr)
+				{
+					entity_to_call = attempted_entity;
+				}
+				else //invalid, take a random contained entity
+				{
+					auto &contained_entities = entity_to_call->GetContainedEntities();
```
domg123-hermes: **Two issues in the invalid-entity branch.**

1. **Null deref (this line).** `entity_to_call` is still `nullptr` here — it's only set in the `attempted_entity != nullptr` branch above. `entity_to_call->GetContainedEntities()` derefs null → segfault (lldb: `Entity::GetContainedEntities(this=0x0)`). It should be `mp.entity`, which is non-null under the enclosing guard and — thanks to the reordered `MutateEntity` (contained entities mutated first) — already holds the new entity's children:
```cpp
auto &contained_entities = mp.entity->GetContainedEntities();
```

2. **Wrong slot (L1741).** The chosen entity id is written to `n_ocn[1]` (the label slot), which leaves the bad reference in `n_ocn[0]` and clobbers the label. It should be `n_ocn[0]`.

With both fixed, `(call_entity "no_such_entity" ...)` repairs to e.g. `(call_entity "Child" ...)` (a real contained entity) instead of crashing.

src/Amalgam/evaluablenode/EvaluableNodeTreeManipulation.cpp
```diff
@@ -1630,16 +1684,91 @@ EvaluableNode *EvaluableNodeTreeManipulation::MutateTree(MutationParameters &mp,
...
+					}
+				}
+			}
 
-	return copy;
+			//only ensure validate label if entity_to_call is valid
+			if(entity_to_call != nullptr && n->GetType() == ENT_CALL_ENTITY)
+			{
+				auto label_sid = EvaluableNode::ToStringIDIfExists(n_ocn[1], true);
+				if(!Entity::IsLabelPrivate(label_sid)
```
domg123-hermes: **Inverted label-privacy logic.** A `call_entity` can only reach a **public** label that exists (`Entity.h:214` returns null for a private label called not-on-self), so invalid = private **or** missing.

- This guard `if(!Entity::IsLabelPrivate(label_sid) || !DoesLabelExist(...))` fires on *public* labels and skips private ones — backwards. Should be `if(Entity::IsLabelPrivate(label_sid) || !entity_to_call->DoesLabelExist(label_sid))`.
- The collection loop at L1757, `if(Entity::IsLabelPrivate(cur_sid)) public_labels.push_back(...)`, gathers *private* labels into `public_labels`. Should be `if(!Entity::IsLabelPrivate(cur_sid))`.

As written, a repaired call is pointed at a private (uncallable) label, while genuinely-invalid private/missing labels are left untouched. Verified A/B on a `C` entity with public `pub` + private `!priv`, calling a nonexistent label: current code → `(call_entity "C" "!priv")`; corrected → `(call_entity "C" "pub")`.

howsoRes (approved):

PR #616, opened 2026-07-17
25799: Fixes bug in mutation when a label is broken

### LINKED ISSUE #615: Mutation call_entity repair still selects private (uncallable) labels
Follow-up to #614 (merged). The review on that PR flagged inverted label-privacy logic in the `call_entity` mutation repair. The **guard** was corrected, but the **label-collection loop** was not — so the repair still cannot produce a valid public label.

### Where
`src/Amalgam/evaluablenode/EvaluableNodeTreeManipulation.cpp`, in `MutateTree`'s `ENT_CALL_ENTITY` repair block (the `public_labels` loop, ~line 1758 on `f0712be`):

```cpp
std::vector<StringInternPool::StringID> public_labels;
for(auto &[cur_sid, cn] : entity_to_call->GetLabelIndex())
{
    if(Entity::IsLabelPrivate(cur_sid))       // collects PRIVATE labels into `public_labels`
        public_labels.push_back(cur_sid);
}
```

### Problem
A `call_entity` can only reach a **public** label — an external (not on-self) call to a private label returns null (`Entity::Execute`, Entity.h). The guard now correctly fires when `n_ocn[1]` is a private-or-missing label, but the loop then gathers **private** labels and assigns one, so the repaired call is pointed at an uncallable private label. If the entity has only public labels, `public_labels` is empty and the bad label is left unchanged. In no case does the repair select a valid public label.

### Repro
```amalgam
(seq
	;Src references contained entity C with a NONEXISTENT label -> repair must pick a valid label.
	(create_entities "Src" (lambda (call_entity "C" "nonexistent_label")))
	;C has one PUBLIC label (pub) and one PRIVATE label (!priv).
	(create_entities (list "Src" "C")
		(lambda
			{
				pub     "public value"
				"!priv" "private value"
			}
		)
	)
	;rate 0.0 so nothing mutates; the call_entity repair still runs unconditionally.
	(mutate_entity "Src" 0.0 "Dst")
	(print "Dst root: " (retrieve_entity_root "Dst") "\n")
)
```

On `f0712be` this prints:
```
Dst root: {
	.null (call_entity "C" "!priv")
}
```
i.e. the repaired call points at the private label. Expected: `(call_entity "C" "pub")`.

### Suggested fix
Collect **public** labels (invert the check):
```diff
-                    if(Entity::IsLabelPrivate(cur_sid))
+                    if(!Entity::IsLabelPrivate(cur_sid))
                         public_labels.push_back(cur_sid);
```
With that change the same repro yields `(call_entity "C" "pub")` (verified locally with a debug build of `f0712be` + this one-line diff).

howsohazard:


howso-mario (approved):
Advisory review verdict: APPROVE.

I reviewed the one-line mutation repair change in `src/Amalgam/evaluablenode/EvaluableNodeTreeManipulation.cpp`. The polarity change is correct: this block is repairing an invalid/private `call_entity` label and should choose from labels where `!Entity::IsLabelPrivate(cur_sid)`, matching the runtime access gate that rejects private labels on other entities.

Evidence checked:
- Diff scope is exactly one source line; `git diff --check origin/main...HEAD` passes and no conflict markers were present in the diff.
- Surrounding `MutateTree` repair logic, `Entity::IsLabelPrivate`, and `call_entity`/entity label access rules were inspected.
- CI is green for the PR build matrix and smoke tests, including Linux amd64/arm64/arm64_8a/WASM, Linux-228, macOS, Windows, and `rebuild-docs / regenerate-docs-files`.
- Local configure/build was attempted with `amd64-release-linux-228` using CMake 4.3.2/Ninja 1.13/GCC 14.2.0. The touched object compiled, but the build later stopped in pre-existing `StringManipulation.cpp` GCC-14 `-Werror=stringop-overflow` behavior documented in the Amalgam review runbook, so I did not treat that local compiler-mismatch failure as a PR regression.
- Independent read-only coding-lane reviews via Claude Code and Codex found no blocking issues.

Non-blocking notes: PR body is empty and there is no focused regression test for the stochastic mutation repair path; a targeted seeded test would improve traceability, but I do not consider it necessary for this one-line bug fix given the CI/build evidence. Also consider adding the expected semver suffix (likely `, PATCH`) to the title before merge if this repo's release-title convention is being enforced manually.

PR #619, opened 2026-07-21
25804: Improves mutate type consistency for call_*_*entity

### LINKED ISSUE #618: Mutation-repair for call_entity doesn't cover the forms native mutation produces (expression entity-slot, missing args)
Follow-up to #614 / #616. Those added a repair pass that fixes *malformed* `call_entity` produced by mutation — an invalid **immediate** entity ref (→ a real contained entity) and, for `ENT_CALL_ENTITY`, a private/missing label (→ a public label). That fixed the crashes and corruption. This is a **capability** follow-up (not a bug, not a regression — everything below is crash-free): can native `mutate_entity` now *generate* a **usable** `call_entity` ADF call — i.e. one that actually dereferences and runs a contained helper?

**Finding: not yet.** The two malformations that dominate what native mutation actually emits are neither covered by the repair, so mutation-produced `call_entity` almost never resolves.

### Repro (cold, self-contained; engine 81.2.1)

Substrate: parent `ind`, body label `eval`, one contained child `h1` with `eval = (+ x 1000)`. Base body `(* x x)` at `x=3` = 9, so any score `>= 1000` proves the body's call ran h1. For each case we score the body directly, then run a **repair-only pass** (`mutate_entity` at rate 0.0 — still runs the `call_entity` cleanup) and score the repaired body:

```amlg
{
  build (declare {body ""}
    (seq
      (if (contains_entity "ind") (destroy_entities "ind"))
      (create_entities "ind" (associate "eval" (parse body)))
      (create_entities (list "ind" "h1") (associate "eval" (parse "(+ x 1000)")))
      "ind"))
  probe (declare {body ""}
    (seq
      (call build {body body})
      (let {direct (call_entity "ind" "eval" (associate "x" 3)
                     {read_access .true write_access .false})}
        (if (contains_entity "mut") (destroy_entities "mut"))
        (mutate_entity "ind" 0.0 "mut" .null .null 1 .null .null)          ; repair-only pass
        (create_entities (list "mut" "h1") (associate "eval" (parse "(+ x 1000)")))
        [direct
         (call_entity "mut" "eval" (associate "x" 3) {read_access .true write_access .false})
         (unparse (retrieve_from_entity "mut" "eval"))])))
  go (declare {} {
     C0_wellformed      (call probe {body "(call_entity \"h1\" \"eval\" {x 3})"})
     C1_expr_entityslot (call probe {body "(call_entity (* x x) \"eval\" {x 3})"})
     C2_missing_args    (call probe {body "(call_entity \"h1\" \"eval\")"})
     C3_bad_label       (call probe {body "(call_entity \"h1\" \"nope\" {x 3})"}) })
}
```

Result on **81.2.1** (`[direct, after-repair, repaired-body]`):

| case | direct | after repair | repair changed it? |
|---|---|---|---|
| C0 well-formed `(call_entity "h1" "eval" {x 3})` | **1003** | 1003 | — (already valid) |
| C1 **expression** in entity slot `(call_entity (* x x) "eval" {x 3})` | null | **null** | **no** |
| C2 **missing args** `(call_entity "h1" "eval")` | null | **null** | **no** |
| C3 bad label `(call_entity "h1" "nope" {x 3})` | null | **1003** | yes → rewrote label to `eval` |

Only C3 — an *immediate* id with a bad label — is repaired. C1 and C2 pass through untouched.

### Why the repair misses them

In `MutateTree`'s `ENT_CALL_ENTITY`/`ENT_CALL_ON_ENTITY` cleanup (`EvaluableNodeTreeManipulation.cpp`):
- The entity-slot branch guards on `EvaluableNode::IsNull(n_ocn[0])` / `n_ocn[0]->IsImmediate()` (~L1721 / L1725). An **expression** in the entity slot is neither, so it's never touched and `entity_to_call` stays null, which also skips the label branch.
- It only ever `n_ocn.resize(2)` (~L1715) — entity + label. It **never synthesizes an args assoc**, so even a perfectly-targeted `(call_entity "h1" "eval")` leaves the callee's params unbound (`x` never binds → h1 returns null).

### How common these forms are

From a chained native-`mutate_entity` probe (weighting `call_entity` into the opcode vocab), of the `call_entity` nodes that referenced h1: ~34% had an immediate-id entity slot, **~50% had an expression** in the entity slot, and **0** carried an args assoc — so **0/520** produced bodies resolved.

### The ask

For native mutation to be able to *build* usable `call_entity` (which would let a GP system let mutation drive ADF/helper-call composition rather than hand-splicing calls), the cleanup would additionally need to: **(a)** when the entity slot is a non-immediate expression, coerce it toward a valid contained-entity id (or drop/replace the node); and **(b)** synthesize a minimal args assoc (even empty) so the target's parameters resolve. Is that in scope for the repair, or intentionally out of it?

Not urgent — crash-free capability gap, well below the crash fixes in priority. Flagging with a runnable repro so the scope question is concrete. (Full harness + form breakdown live in our downstream gplab study `studies/native-call-entity/`.)

howsohazard:


domg123-hermes (commented):
Thanks for taking this on — it addresses both halves of #618. I built the branch at dc97620 and exercised it with small repros; results below.

**What works (confirmed):**
- **Expression entity-slot** now repairs and resolves. `(call_entity (* x x) "eval" {x 3})` → `(call_entity "h1" "eval" {x 3})` → runs the helper. The return-type approach (leave the slot alone when its opcode can return an entity id, coerce only when it can't) is exactly the non-over-constraining shape I was hoping for.
- **Args synthesis** works for arg-free callees: a missing/`.null` args slot becomes `{}`, and a helper that needs no bound vars then resolves. Nice.
- The `AreDataTypesCompatible` → `IsReturnTypeValidForRequiredDataType` rename has **no other call sites**, so the symmetric→directional change is contained to this repair path.

**One blocker + two smaller notes** inline. The blocker (1) is a correctness regression — a *valid* `call_entity` gets its target randomized on every `mutate_entity` — so requesting changes on that; the rest is your call.

Expectation-calibration note (not a defect): synthesizing an **empty** args assoc makes a call well-*formed*, not necessarily semantically live — a helper that needs bound variables (e.g. `(+ x 1000)` wanting `x`) still returns null under `{}`, since the repair can't invent the arguments. That's the right limit; flagging only so "has an args assoc" isn't read as "resolves."

src/Amalgam/evaluablenode/EvaluableNodeTreeManipulation.cpp
```diff
@@ -1711,61 +1711,91 @@ EvaluableNode *EvaluableNodeTreeManipulation::MutateTree(MutationParameters &mp,
...
+			}
 
-				//if valid entity, use it
-				if(attempted_entity != nullptr)
-				{
-					entity_to_call = attempted_entity;
-				}
-				else //invalid, take a random contained entity
+			//if invalid entity and the opcode can't return a list, make it valid
+			if(entity_to_call == nullptr || !can_return_valid_entity)
```
domg123-hermes: **Blocker: this clobbers a *valid* entity reference with a random contained entity on every pass.** `can_return_valid_entity` is only ever set `true` in the expression branch. For a valid **immediate** id (the traverse above succeeds → `entity_to_call != nullptr`), it stays `false`, so this guard is `false || !false` = **true** → the valid target is replaced with a random contained entity. Your own comment says "if invalid entity **and** the opcode can't return a list" — the code is `||`, and reads as `&&`.

Repro (engine built at this head): a valid `(call_entity "h2" "eval" {x 3})` (h2 is a real child with a public `eval`) through a repair-only pass (`mutate_entity` rate 0.0), 40 trials with distinct entity seeds → **17/40 rewritten to a different helper** (`(call_entity "h1" ...)`); the 23 "kept" are just luck (2 children, ~50% land back on h2). So it's not "sometimes preserved" — it re-randomizes every time and the original survives by coincidence. In a GP run this randomizes the target of every valid `call_entity` on each mutation, destroying learned call structure.

Fix — change `||` to `&&`:
```cpp
if(entity_to_call == nullptr && !can_return_valid_entity)
```
Verified: with `&&`, the same repro keeps h2 **40/40**, and the expression-slot case (`(call_entity (* x x) "eval" {x 3})`) still repairs + resolves. All five slot cases (null / valid-immediate / invalid-immediate / expr-can-return-entity / expr-cannot) then behave correctly.

src/Amalgam/evaluablenode/EvaluableNodeTreeManipulation.cpp
```diff
@@ -1711,61 +1711,91 @@ EvaluableNode *EvaluableNodeTreeManipulation::MutateTree(MutationParameters &mp,
...
 					{
-						if(!Entity::IsLabelPrivate(cur_sid))
-							public_labels.push_back(cur_sid);
+						replace_with_valid_label = true;
 					}
+				}
+				else
+				{
+					auto return_types = GetOpcodeReturnTypes(n_ocn[1]->GetType());
+					can_return_valid_entity = OpcodeDetails::IsReturnTypeValidForRequiredDataType(
```
domg123-hermes: **Dead assignment — the expression-label case computes a result nothing reads.** This writes `can_return_valid_entity` (the entity-slot variable) from the label's return-type vs `ENTITY_LABEL`, but after this block only `replace_with_valid_label` is acted on, and that's set solely in the immediate branch above. So the value computed here is never used, and an expression label that *can't* return a valid label is left in place (no symmetric coercion like the entity slot gets).

Probably intended to mirror the entity slot: if the expression can't yield a label, fall into the replace-with-valid-label path. Minor in practice (expression labels are rare in mutation output), but as written it's dead code and slightly misleading (reuses the entity-slot flag name). Worth either wiring it up or dropping it.

src/Amalgam/OpcodeDetails.h
```diff
@@ -190,53 +190,121 @@ class OpcodeDetails
...
+
+		if(AreDataTypesExactlyCompatible(required_type, DataType::ASSOC))
+		{
+			return AreDataTypesExactlyCompatible(return_type, DataType::ASSOC_OF_NUMBERS);
+		}
+
+		if(AreDataTypesExactlyCompatible(required_type, DataType::QUERY))
+		{
+			return (AreDataTypesExactlyCompatible(return_type, DataType::LIST)
+				|| AreDataTypesExactlyCompatible(required_type, DataType::LIST_OF_QUERIES));
```
domg123-hermes: Minor typo in the refactored function (not on the call_entity path, so it doesn't affect this PR's goal): this clause reads `required_type` where every sibling branch reads `return_type`. Since we're already inside `required_type == QUERY`, `AreDataTypesExactlyCompatible(required_type, DataType::LIST_OF_QUERIES)` is statically false, so a `LIST_OF_QUERIES`-returning opcode won't satisfy a `QUERY` requirement.
```cpp
|| AreDataTypesExactlyCompatible(return_type, DataType::LIST_OF_QUERIES));
```

src/Amalgam/evaluablenode/EvaluableNodeTreeManipulation.cpp
domg123-hermes: Addendum after a second pass — the `||` also breaks the **documented null-target form**, and deterministically. `(call_entity .null "me" {})` (null = current entity; the docs even prescribe forming it as `(call_entity .null ...)`) takes this branch too: `entity_to_call = mp.entity` is non-null, but `can_return_valid_entity` is still `false`, so `!can_return_valid_entity` fires. Repair-only pass, 20/20 trials: rewritten to `(call_entity "h1" "eval" {})` — target, label, and args all replaced; the self-call's 777 became the child's 1005. Unlike the immediate-id case there's no coincidental survival — the null form is rewritten **every** time. It's also a regression vs #614, which never replaced a null slot. The same `&&` fix covers it: 20/20 preserved, with the expression-slot repair still working.

For the record, two things from the main review I've since verified rather than assumed: `AreDataTypesExactlyCompatible` is overlap-based (`(a & b) != 0`), so multi-bit return-type masks flow correctly through the new directional checks; and the args-assoc synthesis is valid for `call_on_entity` too (its third param is `assoc params`, and its second — `any function` — is correctly left untouched by the `ENT_CALL_ENTITY`-only label guard).

howsohazard:
Resolved.  Review again.

src/Amalgam/evaluablenode/EvaluableNodeTreeManipulation.cpp
```diff
@@ -1711,61 +1711,91 @@ EvaluableNode *EvaluableNodeTreeManipulation::MutateTree(MutationParameters &mp,
...
 					{
-						if(!Entity::IsLabelPrivate(cur_sid))
-							public_labels.push_back(cur_sid);
+						replace_with_valid_label = true;
 					}
+				}
+				else //dynamic execution, make sure format is appropriate
+				{
+					auto return_types = GetOpcodeReturnTypes(n_ocn[1]->GetType());
+					replace_with_valid_label = OpcodeDetails::IsReturnTypeValidForRequiredDataType(
```
domg123-hermes: **The polarity is inverted — this replaces labels that are fine and skips the ones that need repair.** `IsReturnTypeValidForRequiredDataType(...)` is true when the expression **can** yield a label — but `replace_with_valid_label = <can>` then clobbers exactly the workable case. Mirroring the entity-slot logic, it should replace when the expression **cannot** yield one:
```cpp
replace_with_valid_label = !OpcodeDetails::IsReturnTypeValidForRequiredDataType(
	return_types, OpcodeDetails::DataType::ENTITY_LABEL);
```

Repro at this head (child `h1` has two public labels: `eval` → `(+ x 1000)`, `alt` → `3333`):
- **Working dynamic label**: `(call_entity "h1" (concat "a" "lt") {x 3})` executes fine directly (→ 3333 — the interpreter does evaluate a label expression at runtime). Through a repair-only pass, the `concat` is replaced with a *random* public label **20/20 trials** — so a computed label choice is destroyed and the call's result randomizes between the entity's labels (3333 vs 1003 here).
- **Impossible label**: `(call_entity "h1" (associate "z" 1) {x 3})` (assoc can never be a label) is left untouched and still evaluates null — the case the repair is *for*.

With the `!` added, both flip to correct: computed labels preserved, impossible ones repaired.

domg123-hermes (commented):
Re-reviewed at e386699 — rebuilt and re-ran the full repro set.

**Verified fixed:**
- **`&&` guard**: a valid `(call_entity "h2" "eval" {x 3})` now survives repair-only passes **40/40** (was 17/40 clobbered), and the documented null-target self-call form survives **20/20** (was 20/20 clobbered). The expression-slot repair still works (`(call_entity (* x x) "eval" {x 3})` → repaired → resolves), C0–C3 all behave.
- **QUERY typo**: fixed as suggested.

**One item remains** — the label wire-up (which replaced the dead assignment) has its polarity inverted; detail + repro inline. It's the last thing I can find; with that one flipped this looks done to me.

domg123-hermes:
Verified at d229f53 (rebuilt, full repro battery):

- **Label polarity**: the working dynamic label `(call_entity "h1" (concat "a" "lt") {x 3})` is now preserved **20/20** repair-only passes (still evaluates to 3333), and the impossible label `(call_entity "h1" (associate "z" 1) {x 3})` is now **repaired** to a valid public label and resolves (was left broken). Both directions correct.
- **Regressions**: valid immediate target kept **40/40**; null-target self-call kept **20/20**; C0–C3 all behave (expression entity-slot repairs and resolves; missing args synthesized to `{}`; bad label repaired).

Nothing further from me — every case I can construct now behaves correctly. Thanks for the quick turnarounds.

domg123 (approved):

PR #620, opened 2026-07-27
25798: Improves reliability of docs rebuild pipeline

apbassett:


.github/workflows/rebuild-docs.yml
```diff
@@ -55,13 +55,16 @@ jobs:
...
           if [ -n "$(git status --porcelain)" ]; then
             echo "Changes to docs files detected. Making a new commit..."
             git config user.name "howso-automation"
             git config user.email "support@howso.com"
             git add -A docs/
             git commit -m "Automated docs rebuild"
+            # Reconcile with anything pushed to the branch since checkout so the
+            # push can't fail on a divergent-branch race (default 'git pull' would
+            # abort with "Need to specify how to reconcile divergent branches").
+            git pull --rebase origin "${{ github.head_ref }}"
```
howso-mario: Non-blocking reliability hardening: consider guarding against branch movement instead of rebasing generated docs onto whatever the PR branch tip is at this moment. If a developer push lands while this docs job is running, this can put docs generated from the older artifact on top of a newer code commit; the follow-up bot commit then triggers the docs-bot skip path, so that newer parent may not get its own docs-regeneration run. One safer pattern is to record the checked-out SHA before generation, fetch/compare the remote head before pushing, and exit successfully without pushing if the branch moved (the newer run should regenerate docs from its own artifact).

howso-mario (commented):
Reviewed by Hermes Agent. Verdict: comment / no blocking findings.

Evidence inspected: PR diff and repo guidance; status rollup for run 30305742261 (all build/smoke docs/final-check lanes green, release-only lanes skipped as expected); final-check log shows check-docs-commit, set-pr-version, build, and rebuild-docs all success; rebuild-docs log downloaded the ARM64 artifact and found no generated-doc changes; local `git diff --check origin/main...HEAD` was clean.

The final-check rewrite looks materially safer than the old required-check wrapper because it runs under `always()`, includes set-pr-version/build/rebuild-docs in `needs`, and requires exact success instead of treating skipped dependencies as acceptable. I left one inline non-blocking reliability hardening note for the docs-bot branch-movement window. No merge action taken.

.github/workflows/rebuild-docs.yml
howso-mario: FIXED: Replaced the docs bot `git pull --rebase` path with a branch-movement guard. The workflow now records the checked-out PR SHA before generating docs, fetches the current PR branch head before pushing the generated-doc commit, and exits successfully without pushing if the branch moved so the newer run can regenerate docs from its own artifact.

Verified with `git diff --check`, PyYAML parsing of `.github/workflows/rebuild-docs.yml`, and a read-only Claude Code second-lane review. Pushed as f456598b1e24db4e9a036532a8d39c99342881d4.

howso-mario (approved):
Approved at head f456598b1e24db4e9a036532a8d39c99342881d4. Verified the docs rebuild workflow records the checked-out SHA, fetches and compares the live PR branch head before push, and exits successfully without pushing if the branch moved, addressing the prior reliability note. Inspected current diff and review thread inventory; no blocking findings remain. CI run 30369245810 is complete: required build, cross-platform smoke, rebuild-docs, and final-check lanes passed; release-only lanes were skipped as expected.

PR #622, opened 2026-07-28
25818: Considerable performance improvement for various short string (<= 7 bytes) via pointer tagging

howsohazard:


src/Amalgam/entity/Entity.h
```diff
@@ -401,9 +401,9 @@ class Entity
 		std::vector<EntityWriteListener *> *write_listeners, size_t *num_new_nodes_allocated, bool on_self);
 
 	//Returns the id for this Entity
-	inline const std::string &GetId()
+	inline std::string_view GetId()
 	{
-		return string_intern_pool.GetStringFromID(GetIdStringId());
+		return string_intern_pool.GetStringViewFromID(GetIdStringId());
```
howso-mario: `GetIdStringId()` returns `StringID` by value, so this now passes a temporary into `GetStringViewFromID()`. For an inline (<=7-byte) ID, `GetStringViewFromID()` returns a `string_view` into that `StringID` object's own bytes, and the temporary dies before the caller can use the returned view. That makes common short entity IDs dangle here (e.g. sort/resource/seed paths that consume `GetId()`). Please take the view from the persistent `idStringId` member (for example by returning the `StringID` by `const&`) or otherwise return an owning string, and add a <=7-byte regression test.

src/Amalgam/string/StringInternPool.h
```diff
@@ -19,23 +20,193 @@ class StringInternStringData
...
+		return std::string_view(ptr, InlineStringLength());
+	}
+
+	uint64_t pointerOrShortString;
+};
+
+//add hashing for PointerWithShortInlineString
+namespace std
+{
+	template<class PointerType> struct hash<PointerWithShortInlineString<PointerType>>
```
howso-mario, line 200: This specializes `std::hash`, but the repo containers that store `StringID` use `FastHashMap`/`ConcurrentFastHashMap`, whose default hasher is the repo's `FastHasher`, not `std::hash`. The macOS/Clang CI log fails with `implicit instantiation of undefined template 'FastHasher<PointerWithShortInlineString<StringInternStringData *>>'` from the `staticStringIDToIndex` map. Please add/adjust a `FastHasher` specialization or otherwise make this type satisfy the repo hash-map contract across compilers.

howso-mario (changes requested):
The pointer-tagging direction is promising, but this head is not safe to merge yet.

Blocking issues found:

1. `Entity::GetId()` can return a dangling `std::string_view` for inline/short IDs. The new `GetStringViewFromID` documentation correctly says the `StringID` must be passed by reference because inline views point into the `StringID` object itself, but `GetId()` passes `GetIdStringId()` by value, so <=7-byte IDs return a view into a temporary. This is real UB for common entity ID paths and plausibly explains the entity/smoke regressions.

2. The macOS/Clang CI lane fails compiling `FastHashMap<StringID, ...>` because the PR adds `std::hash<PointerWithShortInlineString<...>>`, while this repo's `FastHashMap` defaults to `FastHasher`. The log shows `implicit instantiation of undefined template 'FastHasher<PointerWithShortInlineString<StringInternStringData *>>'` from `StringInternPool.h:578`.

3. CI-faithful Linux-228 verification compiles but fails runtime validation/smoke tests: 4/13 passed locally and in CI, with both full validation executables failing and every shared-library tester reporting `LoadAndEval produced 0 but expected 24`; the validation also reports `format`, `get_entity_rand_seed`, and `set_entity_rand_seed` failures. Windows shows the same shared-library smoke failure pattern. Other Linux/WASM lanes fail earlier at build stage.

Verification performed:
- Read PR #622 metadata/diff/checks/review threads (no existing reviews/comments/threads found).
- `git diff --check origin/main...HEAD` passed.
- Ran the prescribed CI-faithful container path with `ghcr.io/howsoai/amalgam-build-container-linux-228:2.0.9`, preset `amd64-release-linux-228`: configure passed, all 465 build targets compiled, `cmake --build --preset amd64-release-linux-228 --target test` failed with 9/13 tests failing.
- Checked live GitHub Actions logs for macOS/Linux-228/Windows; results match the local failures above.

Please fix the lifetime issue, add a `FastHasher`-compatible hash path (or otherwise make `StringID` satisfy the repo hash-map contract across compilers), root-cause the validation/smoke regressions, and add focused 0..8-byte boundary/copy-move/hash/equality/serialization tests for the short-string representation.

src/Amalgam/HashMaps.h
```diff
@@ -146,11 +146,17 @@ struct has_data_and_size<T, std::void_t<
...
 inline constexpr bool has_data_and_size_v = has_data_and_size<T>::value;
 
+inline constexpr uint64_t rapid_hasher_rand_seed = 13766731;
+
 //a fast, deterministic hash object
 template <typename T, typename = void>
-struct FastHasher;
-
-inline constexpr uint64_t rapid_hasher_rand_seed = 13766731;
+struct FastHasher
```
howso-mario, line 156: This fixes the Clang compile error by making the *primary* `FastHasher<T>` hash `sizeof(T)` raw bytes for every unsupported type. That silently admits non-trivially-copyable types and types with padding/multiple equal representations, for which equal values need not have equal object bytes (violating the hash/equality contract); it also makes the existing trivially-copyable partial specialization redundant. Please keep the primary undefined/constrained and add a focused `FastHasher<PointerWithShortInlineString<...>>` (or hash an exposed canonical `uint64_t` representation), with cross-compiler equality/hash tests.

howso-mario (dismissed):
Re-review of `4e607b0410514bd9be22384e63c382a9ac49caa3` (one commit after the previously reviewed `95523a0a71be9ae67311e506381abe7e9c39d534`): changes are still required.

What is fixed:
- The short-ID `Entity::GetId()` lifetime bug is fixed by taking the view from persistent `idStringId` storage.
- macOS/Clang now compiles, runs CMake tests, packages, and passes its artifact smoke test.

Remaining blockers:
- Current CI still fails runtime validation on Linux-228/GCC 10.3.1, Linux arm64_8a, and Windows/MSVC. Linux-228 builds cleanly but only 4/13 tests pass: both full-test executables fail `format` tests 33/35 plus `get_entity_rand_seed`/`set_entity_rand_seed`, and all seven shared-library testers return `0` instead of `24` from `(size (contained_entities))`. Windows builds cleanly but fails the same four shared-library tests (12/16 pass). The downstream artifact smoke jobs are skipped. This reproduces the prior failure signature, so the lifetime change was not its root cause.
- The Clang compatibility change makes the unconstrained primary `FastHasher<T>` hash the raw object representation for every otherwise-unsupported type. That is unsafe for non-trivially-copyable types and types with padding or multiple equal representations, and it hides missing hash contracts globally. Please constrain the raw-object path and add a focused `FastHasher<PointerWithShortInlineString<...>>` implementation (or expose a canonical 64-bit representation and hash that), with `equal => equal hash` tests.

Actionable next step for the smoke regression: instrument the `ExecuteEntity("test")` path and label-index lookup. `LoadEntity` succeeds, but after executing the short label `"test"` no contained entities exist, consistently across all library variants. Compare `StringID` raw bits, `IsInlineString()`, length/content, and hashes at label insertion and `GetIDFromString("test")` lookup; then cover 0..8-byte strings (including embedded NUL and high-bit bytes), copy/move, map lookup, serialization, and deterministic entity-random-state behavior under GCC/MSVC/Clang.

Verification: `git diff --check origin/main...HEAD` passed. I read all prior review threads (both are marked resolved); the lifetime thread is fixed, while the FastHasher contract is only partially/supersedingly addressed. A fresh prescribed Linux-228 container configure succeeded and began the 465-target build; I stopped the uncached duplicate build after the exact-head CI job completed with the same preset/compiler and supplied full build/test evidence.

src/Amalgam/HashMaps.h
howsohazard: Added a comment.  I'd rather not over engineer this.

howsoRes (approved):

howso-mario:
Investigation update on the Linux `--validate-amalgam` failure from PR Build `#622 (1.2011)`:

- Target/build: https://github.com/howsoai/amalgam/pull/622, run https://github.com/howsoai/amalgam/actions/runs/30454993590, failing jobs included `build / build-linux-228 (amd64, linux, release)` and `build / build-linux (arm64_8a, linux, release)`.
- Reproduced locally in the CI-faithful `ghcr.io/howsoai/amalgam-build-container-linux-228:2.0.9` image at head `9890acde24b890543f4050e07c210d9eff68026c` (GCC 10.3.1, glibc 2.28): `./out/build/amd64-release-linux-228/amalgam-st --validate-amalgam` failed format tests 33/35 (`date:%S` expected `"45"`, observed empty string; `date:%s` expected `" s"`, observed empty string).
- Root cause appears to be code UB, not the old compiler itself: `StringRef` converts to `StringID` by value, and `GetStringViewFromID(const StringID&)` returns a view into inline `StringID` storage. Passing a `StringRef` directly creates a temporary `StringID`; seven-byte inline type strings like `date:%S`/`date:%s` then leave a dangling `string_view` before the format copy is made.
- Opened focused fix PR: https://github.com/howsoai/amalgam/pull/623
- Verification for #623 in linux-228: focused `amalgam-st --validate-amalgam` -> `All Tests Passed`; full `cmake --build --preset amd64-release-linux-228 -- -j4 && cmake --build --preset amd64-release-linux-228 --target test` -> `100% tests passed, 0 tests failed out of 13`.

This fix should preserve the short-string speedup: it keeps inline IDs in-place and changes only the call sites to take a stable reference to the `StringID` held by `StringRef`; it does not allocate/copy strings on the hot path.

howso-mario:
Update: Claude Code second-lane review of fix PR #623 found the same  -> temporary  view pattern in  random-stream seeding. I added that call site to #623, reran the linux-228 full build/test gate, and it remains green:  -> .

howso-mario:
Correction/update with shell quoting intact: Claude Code second-lane review of fix PR #623 found the same `StringRef` -> temporary `StringID` view pattern in `create_entities` random-stream seeding. I added that call site to #623, reran the linux-228 full build/test gate, and it remains green: `cmake --build --preset amd64-release-linux-228 -- -j4 && cmake --build --preset amd64-release-linux-228 --target test` -> `100% tests passed, 0 tests failed out of 13`.

PR #623, opened 2026-07-29
25818: Fix inline format type string lifetime, PATCH

howso-mario:
## Summary
- Fixes dangling `std::string_view` uses when `StringRef` values holding short inline IDs are passed to `GetStringViewFromID()`.
- Adds `StringRef::GetIDReference()` so call sites can view the stable `StringID` stored inside the `StringRef` instead of a temporary copy.
- Covers the reported `format` type-string path (`date:%S` / `date:%s`) plus the same latent pattern in `create_entities` random-stream seeding.
- Preserves the short-string pointer-tagging fast path: no heap allocation or string copy is added for inline IDs.

## Test Plan
- Reproduced the failure in `ghcr.io/howsoai/amalgam-build-container-linux-228:2.0.9` (GCC 10.3.1, glibc 2.28), at PR #622 head `9890acde24b890543f4050e07c210d9eff68026c`: `./out/build/amd64-release-linux-228/amalgam-st --validate-amalgam` failed format tests 33/35 (`date:%S` expected `"45"`, observed empty string; `date:%s` expected `" s"`, observed empty string).
- After the initial format-path patch in the same container/preset: `cmake --build --preset amd64-release-linux-228 --target amalgam-st-app -- -j2 && ./out/build/amd64-release-linux-228/amalgam-st --validate-amalgam` -> `All Tests Passed`.
- After the second-lane review fix for `create_entities`, in the same container/preset: `cmake --build --preset amd64-release-linux-228 -- -j4 && cmake --build --preset amd64-release-linux-228 --target test` -> `100% tests passed, 0 tests failed out of 13`.

## Review
- Codex identified the root cause and made the fix.
- Claude Code second-lane review confirmed the root cause/fix/performance reasoning and found the additional `create_entities` call site, now fixed in this PR.

### DIFF
--- a/src/Amalgam/interpreter/OpcodesEntityLifecycleAndStorage.cpp
+++ b/src/Amalgam/interpreter/OpcodesEntityLifecycleAndStorage.cpp
@@ -95,7 +95,7 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_CREATE_ENTITIES(EvaluableN
 			continue;
 		}
 
-		auto new_entity_id_string = string_intern_pool.GetStringViewFromID(new_entity_id);
+		auto new_entity_id_string = string_intern_pool.GetStringViewFromID(new_entity_id.GetIDReference());
 		std::string rand_state = entity_container->CreateRandomStreamFromStringAndRand(new_entity_id_string);
 
 		//pause while allocating the new entity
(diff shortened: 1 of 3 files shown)


howsohazard:
Applied a simpler fix in the main branch.

PR #624, opened 2026-07-30
25837: Fixes key string bug, improves mutation operations

howsohazard:
* Fixes key string bug that could cause corrupt JSON output, among other things
* Improves mutation operations to allow for copying nodes anywhere from the tree as opposed to just sibling nodes

### DIFF
--- a/src/Amalgam/evaluablenode/EvaluableNodeTreeManipulation.cpp
+++ b/src/Amalgam/evaluablenode/EvaluableNodeTreeManipulation.cpp
@@ -41,17 +41,17 @@ std::string EvaluableNodeTreeManipulation::MutationParameters::GenerateRandomStr
 	//reuse a string if probable and possible
 	if(rs.Rand() > novel_chance)
 	{
-		if(key_or_symbol_string && keyAndSymbolStrings->size() > 0)
+		if(key_or_symbol_string && valuesFromTree->keyAndSymbolStrings.size() > 0)
 		{
-			size_t rand_index = rs.RandSize(keyAndSymbolStrings->size());
-			return std::string((*keyAndSymbolStrings)[rand_index]);
+			size_t rand_index = rs.RandSize(valuesFromTree->keyAndSymbolStrings.size());
+			return std::string(valuesFromTree->keyAndSymbolStrings[rand_index]);
 		}
 
 		//not a key or symbol string or don't have any, try value string
-		if(valueStrings->size() > 0)
+		if(valuesFromTree->valueStrings.size() > 0)
 		{
-			size_t rand_index = rs.RandSize(valueStrings->size());
-			return std::string((*valueStrings)[rand_index]);
+			size_t rand_index = rs.RandSize(valuesFromTree->valueStrings.size());
+			return std::string(valuesFromTree->valueStrings[rand_index]);
 		}
 	}
 
@@ -69,9 +69,9 @@ std::string EvaluableNodeTreeManipulation::MutationParameters::GenerateRandomStr
 
 	//put the string into the list of considered strings
 	if(key_or_symbol_string)
-		keyAndSymbolStrings->emplace_back(s);
+		valuesFromTree->keyAndSymbolStrings.emplace_back(s);
 	else
-		valueStrings->emplace_back(s);
+		valuesFromTree->valueStrings.emplace_back(s);
 	return s;
 }
 
@@ -547,23 +547,13 @@ EvaluableNode *EvaluableNodeTreeManipulation::MergeTrees(NodesMergeMethod *mm, E
 	return generalized_node;
 }
 
-//used by GetStringsFromTree
-//store strings in raw form; could change it to store ids, but would need to keep
-//a reference of each in case the last node with a string is deleted
-class StringsFromTreeData
-{
-public:
-	std::vector<std::string> keyAndSymbolStrings;
-	std::vector<std::string> valueStrings;
-	EvaluableNode::ReferenceSetType checked;
-};
-
 //returns a set of strings that have appeared in the given tree,
 //separating by strings that appear as keys and symbols by those that are values
-static void GetStringsFromTree(EvaluableNode *tree, StringsFromTreeData &strings_from_tree_data)
+static void GetStringsFromTree(EvaluableNode *tree,
+	EvaluableNodeTreeManipulation::ValuesFromTreeData &strings_from_tree_data)
 {
 	//try to record, but if already checked, then don't do anything
-	auto [_, inserted] = strings_from_tree_data.checked.insert(tree);
+	auto [_, inserted] = strings_from_tree_data.allNodes.insert(tree);
 	if(!inserted)
 		return;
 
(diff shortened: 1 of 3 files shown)


src/Amalgam/evaluablenode/EvaluableNodeTreeManipulation.cpp
```diff
@@ -1404,22 +1394,35 @@ EvaluableNode *EvaluableNodeTreeManipulation::MutateNode(EvaluableNode *n, Mutat
...
+
 		//copy one element over another
 		if(n->GetOrderedChildNodes().size() > 0)
 		{
 			//get source and destination; note that destination_index is drawn from
 			//num_children - 1 to give a different index, but then includes appending to the end / new value
 			//so it gets a + 1 added back, and then is incremented if greater or equal to source_index
 			size_t num_children = n->GetOrderedChildNodesReference().size();
-			size_t source_index = mp.interpreter->randomStream.RandSize(num_children);
 			size_t destination_index = mp.interpreter->randomStream.RandSize(num_children);
```
howso-mario: `RandSize(num_children)` only returns values in `[0, num_children)`, so `destination_index >= num_children` below can never be true. The same problem affects the mapped-container new-key branch at line 1437. This removes the growth behavior that `replace_element_with_copy` previously had (append/add-new-key), while the comments still describe that outcome. Please preserve the `num_children + 1` destination space (or otherwise explicitly keep the append/new-key outcome) and add focused ordered/mapped mutation tests.

src/Amalgam/evaluablenode/EvaluableNodeTreeManipulation.cpp
```diff
@@ -1404,22 +1394,35 @@ EvaluableNode *EvaluableNodeTreeManipulation::MutateNode(EvaluableNode *n, Mutat
 		break;
 
 	case ENBISI_replace_element_with_copy:
+	{
+		//select a source from all the nodes in the original tree
+		size_t source_index = mp.interpreter->randomStream.RandSize(mp.valuesFromTree->allNodes.size());
+		EvaluableNode *source_node = nullptr;
+		//iterate over child nodes until find the right source index
+		size_t cur_index = 0;
+		for(auto &cn : mp.valuesFromTree->allNodes)
```
howso-mario, line 1403: This selects the source by iterating `allNodes`, which is a pointer-keyed hash set. That maps the random index to nodes according to pointer addresses and hash-container iteration order, so fixed-seed mutation can vary across runs/builds/platforms. Please keep a separate deterministic traversal-order vector for source selection (using the set only for cycle/visited detection) and cover fixed-seed reproducibility.

howso-mario (dismissed):
## Hermes advisory review: Changes requested

Reviewed PR #624 at head `848e28065cc9dfdf1058d8b8e16988dfd6764037` against `origin/main` (`b10ad91f105b23a95cd8455ad303febb31f00cf9`).

The `Parser::GetUnparseStringFromKey(const StringID&)` change looks like the right lifetime fix for the short-inline key JSON corruption path in the current lvalue callers, and the PR builds/tests cleanly. However, the broadened mutation-source change introduces two correctness/reproducibility blockers in `replace_element_with_copy`; see inline comments.

Verification performed:
- PR metadata/diff/check state/review threads inspected; no existing review threads were present.
- Repo docs read: `README.md`, `CONTRIBUTING.md`; no `AGENTS.md` or repo-native `.claude`/rules files found.
- `gh pr checks 624`: all required PR Build lanes green, including Linux amd64/arm64/arm64_8a/WASM, Linux glibc-2.28 amd64, macOS arm64, Windows amd64, smoke tests, `rebuild-docs / regenerate-docs-files`, and `final-check`.
- Local CI-faithful container gate: `ghcr.io/howsoai/amalgam-build-container-linux-228:2.0.9`, preset `amd64-release-linux-228`, `cmake --preset`, full `cmake --build --preset`, and `cmake --build --preset --target test` completed; CTest reported `100% tests passed, 0 tests failed out of 13`.
- `git diff --check origin/main...HEAD` fails on trailing whitespace at `src/Amalgam/evaluablenode/EvaluableNodeTreeManipulation.cpp:1434`.

Residual risk after the blockers are fixed: add focused regression coverage for short inline JSON keys (sorted and unsorted), ordered/mapped mutation append/new-key behavior, deterministic fixed-seed mutation, and deep-copy/no-alias behavior when copying from outside the destination sibling set.

src/Amalgam/evaluablenode/EvaluableNodeTreeManipulation.cpp
howsohazard: That would incur unwanted performance overhead and would not improve determinism because it is the same sort of structure used when iterating over mapped nodes.  Leaving as-is.

src/Amalgam/evaluablenode/EvaluableNodeTreeManipulation.cpp
howsohazard: Resolved.  Testing mutations is beyond the scope of unit tests and is done in a separate repository, evolver.

howsoRes (approved):

src/Amalgam/evaluablenode/EvaluableNodeTreeManipulation.cpp
```diff
@@ -1404,49 +1394,44 @@ EvaluableNode *EvaluableNodeTreeManipulation::MutateNode(EvaluableNode *n, Mutat
...
-			if(destination_index >= source_index)
-				destination_index++;
+			size_t destination_index = mp.interpreter->randomStream.RandSize(num_children + 1);
 
 			if(destination_index >= num_children)
-				n->AppendOrderedChildNode(mp.enm->DeepAllocCopy(n->GetOrderedChildNodes()[source_index]));
+				n->AppendOrderedChildNode(mp.enm->DeepAllocCopy(source_node));
 			else
-				n = n->GetOrderedChildNodes()[destination_index] = mp.enm->DeepAllocCopy(n->GetOrderedChildNodes()[source_index]);
+				n = n->GetOrderedChildNodes()[destination_index] = mp.enm->DeepAllocCopy(source_node);
```
howso-mario, line 1424: This still returns the copied child instead of the mutated container on the ordered-overwrite path. Because `MutateTree` uses `n = MutateNode(n, mp, depth)`, this assignment replaces the whole current subtree with `source_node` whenever `destination_index < num_children`; the write into `GetOrderedChildNodes()[destination_index]` is then applied to an orphaned container. The append and mapped-overwrite paths keep returning the container. Please drop the leading `n =` so this operation only overwrites the selected child slot.

howso-mario (changes requested):
Thanks for the follow-up. I re-reviewed head 2116bf224b5056df261fee6473b9625b914e7afe. The original append/new-key off-by-one is partly addressed: both ordered and mapped branches now draw from num_children + 1, so the append/new-key path is reachable again. However, there is still a blocking ordered-overwrite bug: the overwrite path assigns the copied child into the slot and also into n, so MutateNode returns the child copy and replaces the whole current subtree instead of preserving the container.

On the two replies: I can accept keeping mutation coverage in the separate evolver repo as a project testing policy, though this would benefit from a focused deterministic regression. I do not think the deterministic-source won't-fix rationale is technically convincing: source selection still iterates a pointer-keyed hash set, so fixed-seed behavior can depend on allocation/hash iteration order; a companion traversal-order vector would add one amortized O(1) push per newly visited node and would make source selection O(1) instead of scanning the hash set per replacement. That is not the blocker in this review, but I recommend revisiting it.

CI/checks inspected: current PR checks are green including final-check, linux/mac/windows/wasm builds and smoke tests, rebuild-docs, check-docs-commit, and set-pr-version. Locally, git diff --check still reports trailing whitespace at EvaluableNodeTreeManipulation.cpp:1431, and a CI-faithful linux-228 container build progressed through configuration and 403/465 build steps before the 600s worker timeout, with no compile error observed before timeout.

Verdict: changes requested for the ordered-overwrite whole-subtree replacement below; after that is fixed, the deterministic source selection is the remaining risk/maintainer judgment call.

PR #625, opened 2026-07-30
25824: Fixes deep memory issue, improves performance for accum, improves performance for string interning, allows accum_to_entities to accum to non-existent labels, fix for mix, MINOR

howsohazard:
* The documentation stated that accum_to_entities should have created non-existent entities.  This corrects that.
* This PR improves performance of accum in many workflows especially with hybrid new and not-new references.
* It also fixes an issue that could cause inconsistent flags in entities' data, where a node may have cycles but not update EvaluableNode flags upstream.  This incurs extra copying which decreases performance and has significant cross-cutting changes.
* It fixes an issue with mutate* opcodes when grafting via replace_element_with_copy would not append properly.
* The extra copying above is mitigated by an improvement to refcounting for string interning.


src/Amalgam/string/StringInternPool.h
```diff
@@ -449,30 +461,22 @@ class StringInternPool
...
-			//if can decrement, return
-			//release order on the decrement, don't need ordering on the failure path
-			if(sd_ptr->refCount.compare_exchange_weak(ref_count, ref_count - 1,
-				std::memory_order_release,
-				std::memory_order_relaxed))
-				return;
-		}
+
+		//decrement counter
+		size_t prev = sd_ptr->refCount.fetch_sub(1, std::memory_order_acq_rel);
```
howso-mario: Blocking: this now drops the last reference (`1 -> 0`) before holding the shard lock. That reopens the race described by the deleted comment: thread A can `fetch_sub` to 0 and pause; thread B can intern the same string by finding the still-present map entry, increment 0 -> 1, then release it and also enter the erase path. If B erases first, A later dereferences `sd_ptr->stringData` after the `unique_ptr` was destroyed; if `find` returns `end()`, the subsequent `erase(iterator_with_lock)` is also unsafe. The final transition to zero needs to happen while holding the shard lock (or use a real deletion/tombstone state); checking `iterator_with_lock == end()` would not by itself fix the UAF before/during `find`.

src/Amalgam/evaluablenode/EvaluableNode.h
```diff
@@ -966,6 +970,93 @@ class EvaluableNode
 	}
 #endif
 
+	//returns true if the top node has never been read / accessed
+	__forceinline bool GetIsFreeableTopNode()
+	{
+		return HasAttribute(Attribute::FREEABLE);
```
howso-mario: Blocking/simple correctness bug: `GetIsFreeableTopNode()` is testing `Attribute::FREEABLE`, while the setter and atomic getter use `Attribute::FREEABLE_TOP_NODE`. Non-atomic callers therefore read a different bit than the one this API sets, which can corrupt the new top-node freeability/uniqueness protocol used by the assign/accum changes.

src/Amalgam/interpreter/OpcodesVariableDefinitionAndModification.cpp
```diff
@@ -748,24 +774,24 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_ASSIGN_and_ACCUM(Evaluable
...
 #endif
 
 	//if writing to an outer scope, can't guarantee the memory at this scope can be freed
-	any_nonunique_assignments |= !top_of_stack;
+	any_nonunique_assignments |= !symbol_location.atTopOfStack;
 
 	//make a copy of value_replacement because not sure where else it may be used
 	EvaluableNode *value_replacement = nullptr;
-	if(*value_destination == nullptr)
+	if(symbol_location.location == nullptr)
```
howso-mario: This should still appear to test the stored value (`*symbol_location.location == nullptr`), not the slot pointer. `GetScopeStackSymbolLocation(..., create_if_nonexistent=true, ...)` returns a pointer to the map slot, so `location` itself should be non-null even for a fresh/null variable. As written the `ENT_NULL` seed branch is unreachable and a null `value_replacement` can flow into `TraverseToDestinationFromTraversalPathList`, whereas the previous code always seeded traversal from an `ENT_NULL` node.

howso-mario (changes requested):
Code-inspection-only review, focused on the `StringInternPool` atomics and the related assign/accum lifetime changes. I did not run local builds or tests by design per the review request.

Verdict: changes requested. I found a blocking `StringInternPool` last-reference race that can reintroduce rare heap corruption under `MULTITHREAD_SUPPORT`, plus two nearby copy/lifetime/freeability issues that look like straightforward correctness bugs. The overall direction makes sense, but I don't think this head is safe to merge until these are addressed.

src/Amalgam/string/StringInternPool.h
```diff
@@ -449,28 +461,17 @@ class StringInternPool
...
-			//release order on the decrement, don't need ordering on the failure path
-			if(sd_ptr->refCount.compare_exchange_weak(ref_count, ref_count - 1,
-				std::memory_order_release,
-				std::memory_order_relaxed))
-				return;
-		}
 
-		//lock this shard and double-check that it's the last reference before erasing
+		//decrement counter
+		size_t prev = sd_ptr->refCount.fetch_sub(1, std::memory_order_acq_rel);
```
howso-mario: Blocking: the race still exists. The undo-increment happens too late because after this `fetch_sub` returns `1`, this thread no longer owns `sd_ptr` and is not holding the shard lock yet. A possible interleaving is: T1 destroys the last ref and does `1 -> 0`, then pauses before the undo; T2 creates the same string by content, finds the still-present map entry, increments `0 -> 1`, then destroys it and erases the map entry under the shard lock; T1 resumes and writes/reads freed memory at the undo or `find(sd_ptr->stringData)`.

The safe invariant is that `refCount` must not reach zero outside the shard lock unless creators are changed to treat zero as a tombstone. A closer-to-current, bounded fast path is a single CAS attempt that only decrements when the observed count is greater than one; on contention or possible last-ref it falls back to the locked path, with no CAS retry loop:

```cpp
auto sd_ptr = id.GetPointer();
#if defined(MULTITHREAD_SUPPORT)
	// Fast path: decrement only if this cannot be the last reference.
	// The count must never reach zero outside the shard lock, because a
	// concurrent create-by-string can otherwise resurrect the map entry and a
	// later destroy can free sd_ptr before this thread reaches the lock.
	size_t cur = sd_ptr->refCount.load(std::memory_order_relaxed);
	if(cur > 1 && sd_ptr->refCount.compare_exchange_strong(
			cur, cur - 1,
			std::memory_order_release,
			std::memory_order_relaxed))
		return;

	// Last reference, or a rare race lost by the fast path. This thread
	// still owns its reference here, so sd_ptr/stringData remain alive while
	// acquiring the shard lock.
	auto iterator_with_lock = stringToID.find(sd_ptr->stringData);
	if(iterator_with_lock == stringToID.end())
	{
		AmlgAssert(false);
		return;
	}

	size_t prev = sd_ptr->refCount.fetch_sub(1, std::memory_order_acq_rel);
	if(prev > 1)
		return;

	stringToID.erase(iterator_with_lock);
#else
```

This keeps the common non-final decrement to one atomic RMW with no retry loop. If another thread changes the count between the relaxed load and the CAS, the code just pays the shard-lock fallback, which is correct and bounded. It also makes `find(sd_ptr->stringData)` safe because this thread has not released its reference before taking the lock. Optional extra hardening: assert/guard that `iterator_with_lock != stringToID.end()` before `erase(iterator_with_lock)`.

howso-mario (changes requested):
Follow-up code-inspection-only review of the `StringInternPool` atomic fix at head `d38c7d36d23744e991a06a96af4e129311d7f05e`. I still think there is a blocking race in the last-reference path; see the inline comment for the interleaving and a concrete replacement. I did not run builds/tests per the focused review request.

howso-mario:
Follow-up on the signed `int64_t` / negative-refcount idea at head `d38c7d36d23744e991a06a96af4e129311d7f05e`:

I think the signed-negative approach can be made correct, and it is probably the lowest-overhead version for the destroy path, but only if the refcount is redefined as a small state machine rather than just changing the type.

Recommended invariant:

- Use `std::atomic<int64_t> refCount` as “extra references beyond the first”.
  - `refCount >= 0` means the entry is live with `refCount + 1` references.
  - `refCount == -1` means deletion has been claimed. This is a terminal tombstone state until the map entry is erased.
- New dynamic strings initialize `refCount` to `0`, not `1`.
- Destroy is allowed to do `0 -> -1` outside the shard lock because `-1` blocks resurrection. Exactly one destroyer can observe `prev == 0`; `prev < 0` should assert as over-release/double-destroy.
- No create path may transition `-1` back to a live value. String-keyed creation that finds a tombstoned entry must release the shard lock and retry after the claimant erases it.
- Guard `INT64_MAX` on increments; do not rely on signed wrap. Reporting/debug code such as `GetDynamicStringsInUse()` should report `refCount + 1` for live entries.

Sketch:

```cpp
class StringInternStringData
{
public:
    template<class T, class = std::enable_if_t<std::is_constructible_v<std::string, std::decay_t<T>>>>
    StringInternStringData(T &&s)
        : refCount(0), stringData(std::forward<T>(s))
    {}

#if defined(MULTITHREAD_SUPPORT)
    std::atomic<int64_t> refCount;
#else
    int64_t refCount;
#endif
    std::string stringData;
};

#if defined(MULTITHREAD_SUPPORT)
static inline bool TryAddStringRef(StringInternStringData *sd_ptr)
{
    int64_t count = sd_ptr->refCount.load(std::memory_order_relaxed);
    while(count >= 0)
    {
        AmlgAssert(count < INT64_MAX);
        if(sd_ptr->refCount.compare_exchange_weak(
            count, count + 1,
            std::memory_order_acquire,
            std::memory_order_relaxed))
            return true;
    }

    return false; // deletion already claimed; caller must not resurrect
}
#endif
```

For `CreateStringReference(const std::string&)` and the `std::string_view` overload, use that helper while the `ConcurrentFastHashMap` iterator holds the shard lock:

```cpp
while(true)
{
    auto inserted = stringToID.emplace(str, nullptr);
    if(inserted.second)
    {
        inserted.first->second = std::make_unique<StringInternStringData>(str);
        return StringID(inserted.first->second.get());
    }

#if defined(MULTITHREAD_SUPPORT)
    if(TryAddStringRef(inserted.first->second.get()))
        return StringID(inserted.first->second.get());

    // Found a tombstoned entry (`refCount == -1`). Let the deleter acquire
    // this shard lock and erase it, then retry and either find the fresh entry
    // or insert a new one. Do not increment -1 back to 0.
    continue;
#else
    inserted.first->second->refCount++;
    return StringID(inserted.first->second.get());
#endif
}
```

And `DestroyStringReference` becomes one refcount RMW on both the non-final and final paths:

```cpp
auto sd_ptr = id.GetPointer();

#if defined(MULTITHREAD_SUPPORT)
int64_t prev = sd_ptr->refCount.fetch_sub(1, std::memory_order_acq_rel);
if(prev > 0)
    return;

AmlgAssert(prev == 0); // prev < 0 means an over-release/double destroy

// We uniquely claimed deletion by moving 0 -> -1. Since creators must not
// resurrect -1, sd_ptr remains live until this claimant erases the map entry.
auto iterator_with_lock = stringToID.find(sd_ptr->stringData);
AmlgAssert(iterator_with_lock != stringToID.end());
AmlgAssert(iterator_with_lock->second.get() == sd_ptr);
stringToID.erase(iterator_with_lock);
#else
int64_t prev = sd_ptr->refCount--;
if(prev > 0)
    return;
AmlgAssert(prev == 0);
stringToID.erase(sd_ptr->stringData);
#endif
```

For the `StringID`/bulk add-ref helpers, the safest version is to use the same non-negative CAS helper and assert success; under their current contract the caller already owns/guarantees a live ID, so seeing `-1` would indicate a caller-side lifetime race. If you keep those as blind `fetch_add`, the sentinel invariant relies entirely on that precondition and a misuse could resurrect a tombstone.

Comparison:

- Current branch: non-final destroy is one RMW, but final destroy does `fetch_sub` + undo `fetch_add` + locked `fetch_sub`; worse, the temporary zero still permits the same resurrection/UAF window before the undo.
- Previous CAS recommendation: keeps correctness by never publishing zero outside the lock; common destroy is load + successful CAS, with retries only under real contention. It is simpler and still a safe fallback.
- Signed tombstone scheme: common destroy is a single unconditional `fetch_sub`; final destroy is also a single `fetch_sub` plus shard lock/erase. The cost shifts to string-keyed creation, which must use load/CAS and the rare `-1` retry instead of blind `fetch_add`.

So my recommendation is: yes, the signed negative scheme is viable and likely the best performance/correctness tradeoff if the above tombstone invariant is applied consistently. If that feels too invasive, the lowest-overhead safe fallback remains the bounded CAS destroy path from the prior comment; what is not safe is letting the count become zero outside the shard lock while creators can still increment that same map entry.

howso-mario (approved):
Final rereview at head `839194531dd556f51114ca6109302925e6e3dc08`.

Verdict: merge-ready from the requested code-inspection scope. I did not find remaining blockers in the updated `StringInternPool` single-CAS path or in the related assign/accum/entity correctness changes.

Reasoning:
- The current `DestroyStringReference` fast path only performs an unlocked decrement when the observed count is `> 1` and the single CAS succeeds. It therefore does not transition the last reference to zero outside the shard lock.
- If the CAS fails, or the observed count may be the last reference, it falls back to `stringToID.find(sd_ptr->stringData)`. In the `ConcurrentFastHashMap` implementation this iterator owns the shard lock, so string-based creation cannot increment/resurrect the same map entry between the final decrement and erase.
- Under that shard lock, `fetch_sub` either observes `> 1` and leaves a live entry, or observes the last reference and erases while holding the same shard lock. I do not see the stale-reference/resurrection interleaving from the prior reviews in this version.
- Omitting the extra assertion branch in this hot path is acceptable to me. A debug/integrity-only assertion that the locked decrement never observes zero could be useful diagnostics for caller-contract violations, but it is not needed for the valid-ownership algorithm and I would not block merge over its absence.
- The prior `FREEABLE_TOP_NODE` getter issue is fixed: the regular and atomic accessors now read the `FREEABLE_TOP_NODE` bit.
- The prior `assign`/`accum` destination check is fixed: the code now tests `*symbol_location.location` rather than the slot pointer, so newly-created null slots are handled as null values.
- The `assign_to_entities`/`accum_to_entities` interaction now intentionally allows accumulation to missing labels by creating the label, consistent with the updated opcode text/example; I did not find a correctness regression in the inspected ownership/copy paths.

Residual risk: this rereview was intentionally code-inspection-only per request. I did not run a broad local build or test suite; several PR CI jobs were still in progress when I reviewed, while the completed visible checks I inspected were green.

PR #634, opened 2026-08-10
25662: Adds simplify opcode, improves consistency of subtraction and division opcodes, MINOR

howsohazard:
* Adds simplify opcode that performs basic simplifications
* Unitary division now performs reciprocal analogous to unitary minus, and pow now supports unitary operation like addition and multiplication
* Fixes numerous consistency issues regarding code mixing and merging, especially involving multiplication and division
* Enhances flexibility of a number of math opcodes


src/Amalgam/evaluablenode/EvaluableNodeTreeAlgebra.cpp
```diff
@@ -0,0 +1,785 @@
...
+			EvaluableNodeReference result = interpreter->InterpretNode(en);
+			if(result.unique && nodes_freeable)
+			{
+				for(auto &cn : en->GetOrderedChildNodes())
+					enm->FreeNodeTree(cn);
+				for(auto &[_, cn] : en->GetMappedChildNodes())
+					enm->FreeNodeTree(cn);
+			}
+			en->CopyValueFrom(result);
+			en->CopyMetadataFrom(result);
```
howso-mario: Blocking: `CopyMetadataFrom` dereferences its argument, but `InterpretNode` legitimately represents null with a null pointer. `(simplify (lambda (-)))` and `(simplify (lambda (/)))` both SIGSEGV here (exit 139), including ST, AFMI, and PGC builds. `CopyValueFrom` on the preceding line already handles `nullptr`, so the low-effort fix is `if(result == nullptr) en->ClearMetadata(); else en->CopyMetadataFrom(result);`, plus regressions for both zero-argument expressions.

src/Amalgam/evaluablenode/EvaluableNodeTreeAlgebra.cpp
```diff
@@ -0,0 +1,785 @@
...
+					enm->FreeNodeTree(ocn[i]);
+				ocn.erase(begin(ocn) + i);
+
+				//recheck this position next iteration
+				i--;
+			}
+			else //result is false
+			{
+				short_circuit = true;
+				short_circuit_value = false;
```
howso-mario: Blocking: the absorbing value is hard-coded to `false`, which is correct for `(and ... .false)` but wrong for `(or ... .true)`. `(call (simplify (lambda (or a .true))) {a .false})` returns `.false`; the original returns `.true`. This assignment should use the operator's absorbing value, e.g. `short_circuit_value = is_or`, with paired `or/.true` and `and/.false` regressions.

src/Amalgam/evaluablenode/EvaluableNodeTreeAlgebra.cpp
```diff
@@ -0,0 +1,785 @@
...
+
+			bool immediate_value = EvaluableNode::ToBool(ocn[i]);
+
+			//eliminate or short-circuit based on value
+			if((!is_or && immediate_value) ||  (is_or && !immediate_value))
+			{
+				//erase the node
+				if(nodes_freeable)
+					enm->FreeNodeTree(ocn[i]);
+				ocn.erase(begin(ocn) + i);
```
howso-mario, line 230: Blocking: `and` returns `conditionN`, not merely its truthiness, so erasing a trailing neutral immediate changes the result. `(call (lambda (and a .true)) {a 3})` returns `.true`, while the simplified form returns `3`. For `ENT_AND`, preserve a neutral immediate when it is the final child (or otherwise preserve the value-determining operand); add a regression that compares both value and type. `or` does not have this trailing-value issue because false operands never determine its result.

src/Amalgam/evaluablenode/EvaluableNodeTreeAlgebra.cpp
```diff
@@ -0,0 +1,785 @@
...
+			currently_accumulating_nonimmediate = false;
+			nonimmediate_accum_multiplicand = 0.0;
+		}
+
+		//add on accumulated_immediate at end if nonzero
+		if(accumulated_immediate != 0.0)
+		{
+			if(ocn.size() > 0)
+			{
+				ocn.push_back(enm->AllocNode(subtraction ? -accumulated_immediate : accumulated_immediate));
```
howso-mario: Blocking: the subtraction fold does not preserve the distinguished first operand or the new unary contract. A leading immediate is erased and the surviving symbol is promoted to `x1`: `(- 5 a)` simplifies to `(- a -5)` (at `a=2`, `3` becomes `7`). Separately, when the accumulator is zero, `(- a 0)` and `(- a 5 -5)` collapse to unary `(- a)`, changing `a` into `-a`. For `ENT_SUBTRACT`, fold only denominator/tail immediates unless the whole expression collapses, and if one child would remain either replace the opcode with that child or retain the zero identity. Please cover leading-immediate and identity-cancellation cases.

src/Amalgam/evaluablenode/EvaluableNodeTreeAlgebra.cpp
```diff
@@ -0,0 +1,785 @@
...
+			currently_accumulating_nonimmediate = false;
+			nonimmediate_accum_exponent = 1.0;
+		}
+
+		//add on accumulated_immediate at end if nonzero
+		if(product_immediate != 1.0)
+		{
+			if(ocn.size() > 0)
+			{
+				ocn.push_back(enm->AllocNode(division ? -product_immediate : product_immediate));
```
howso-mario: Blocking: this division fold has three semantic failures. (1) `product_immediate` is the accumulated quotient, so `(/ a 2)` becomes `(/ a -0.5)` (at `a=2`, `1` becomes `-4`). (2) Erasing a leading immediate promotes a denominator into `x1`: `(/ 6 a)` becomes `(/ a -6)` (at `a=2`, `3` becomes `-1/3`). (3) an identity accumulator drops the last child, so `(/ a 1)` and `(/ a 2 0.5)` become unary `(/ a)`, changing `a` into `1/a`. A one-line reciprocal/sign fix is insufficient. Please preserve the first-position role and prevent arity collapse; normalizing symbolic denominator terms to multiplication by reciprocal powers is a robust option. Add equivalence tests for all three shapes plus zero/infinity edges.

howso-mario (dismissed):
Blocking correctness issues remain in the new `simplify` behavior: a null constant result can crash, boolean rewrites can change returned values, and subtraction/division folding can change semantics (including common identity cases introduced by the new unary contracts). See the focused inline comments.

Verification: CI-faithful `ghcr.io/howsoai/amalgam-build-container-linux-228:2.0.9` configure/build/test completed all 472 build steps and passed 13/13 CTest smoke targets (including 64/64 built-in validation tests and 67/67 shared-library checks). Focused runtime probes reproduced every inline issue on ST; the null-result crash also reproduced on AFMI and PGC. Latest head is a generated-doc-only commit; its PR run is green with build/docs jobs skipped by the docs-commit gate, while the immediately preceding source-head run completed all product build/smoke/docs-regeneration jobs successfully before being marked cancelled by the head update. These failures therefore expose missing focused regression coverage rather than build/CI breakage.

src/Amalgam/Merger.h
```diff
@@ -232,6 +232,24 @@ class Merger
 			unmatched_a1.reserve(a1.size());
 		}
 
+		//first, find and merge exact matches
+		for(size_t i1 = 0; i1 < a1.size(); i1++)
+		{
+			for(size_t i2 = 0; i2 < a2.size(); i2++)
+			{
+				if(EvaluableNode::AreDeepEqual(a1[i1], a2[i2]))
```
howso-mario: Non-blocking, low-effort maintainability suggestion: this generic base template now hard-codes `EvaluableNode::AreDeepEqual`, and `Merger.h` does not include `EvaluableNode.h`, so it acquires a fragile include-order dependency and prevents other `Merger<T>` specializations from using `MergeUnorderedSets`. No current non-node specialization calls this method, so it is latent rather than a present build break. Please move exact-match testing behind a virtual/type-specific hook, or use the metric contract with `auto r = MergeMetric(...); if(r.exactMatch && r.IsNontrivialMatch())` (the nontrivial guard matters because default metric results initialize `exactMatch=true`).

src/Amalgam/evaluablenode/EvaluableNodeTreeAlgebra.cpp
howsohazard: Resolved in CopyMetadataFrom to guard against this crash in other places.

src/Amalgam/evaluablenode/EvaluableNodeTreeAlgebra.cpp
howsohazard: Resolved by describing in the documentation for simplify.

src/Amalgam/evaluablenode/EvaluableNodeTreeAlgebra.cpp
howso-mario: Fixed in ef33c69. SUBTRACT is excluded from simplifier regrouping and sorting, preserving the distinguished first operand, arity, and the exact operand sequence present on entry. Focused regressions cover a leading immediate, zero identity, more-than-two operands, and a rounding-sensitive value-equivalence case.

src/Amalgam/evaluablenode/EvaluableNodeTreeAlgebra.cpp
howso-mario: Fixed in ef33c69. DIVIDE is excluded from simplifier regrouping and sorting, preserving the distinguished numerator, arity, and the exact operand sequence present on entry. Focused regressions cover the three reported shapes, more-than-two operands, zero/infinity/signed cases, and an underflow-sensitive value-equivalence case.

src/Amalgam/evaluablenode/EvaluableNodeTreeAlgebra.cpp
howso-mario: Follow-up: Christopher clarified that IEEE operation-order preservation is not a simplify constraint, so the ef33c69 blanket exclusion is superseded. 02d22ce restores canonical SUBTRACT sorting/folding while preserving the distinguished minuend and runtime arity: leading immediates remain in position, binary identities use unary `(* a)` as a numeric carrier, and head cancellation/sign cases (`a-a`, `a-a-a`, `a-a-b`) are covered. Null and no-interpreter empty-arity regressions are also included.

src/Amalgam/evaluablenode/EvaluableNodeTreeAlgebra.cpp
howso-mario: Follow-up: Christopher clarified that IEEE operation-order preservation is not a simplify constraint, so the ef33c69 blanket exclusion is superseded. 02d22ce restores canonical DIVIDE sorting/folding with the reciprocal fix while preserving the distinguished dividend and runtime arity. Regressions cover leading immediates, `/ a 1`, grouped denominators (`/ b a a 1 3 a`), head cancellation/exponents (`a/a`, `a/a/a`, `a/a/b`), zero-divisor signs, null children, and no-interpreter empty arity.

src/Amalgam/evaluablenode/EvaluableNodeTreeAlgebra.cpp
```diff
@@ -0,0 +1,999 @@
...
+
+				//need to recheck index since erased one
+				i--;
+			}
+		}
+
+		//append all at the end
+		if(associative_child_nodes.size() != 0)
+		{
+			en->AppendOrderedChildNodes(associative_child_nodes);
```
howso-mario: Blocking: flattening appends nested children at the end, which permutes operands for associative-but-ordered `concat`. Current HEAD reproducer:

```amlg
(call (simplify (lambda (concat "a" (concat b "c") "d"))) {b "X"})
```

returns `"adXc"`; the original returns `"aXcd"`. `ENT_CONCAT` is `ORDERED` and `isAssociative`, so associativity permits removing parentheses but not reordering operands. Please splice each nested node's children at the erased child's original index (preserving their order) instead of collecting and appending them. Add focused simplify/call examples that assert `(concat "a" b "cd")` and `"aXcd"`, ideally including two separated nested concats.

src/Amalgam/Parser.cpp
```diff
@@ -1267,7 +1273,8 @@ void Parser::Unparse(UnparseData &upd, EvaluableNode *tree, EvaluableNode *paren
 				//move current code back in
 				std::swap(previous_code, upd.result);
 
-				std::sort(begin(unordered_code), end(unordered_code), StringManipulation::StringNaturalCompareSort);
+				std::sort(begin(unordered_code) + starting_index, end(unordered_code),
```
howso-mario: Blocking safety issue: the new `ONE_POSITION_THEN_UNORDERED_OR_ONE_UNORDERED` schema sets `starting_index = 1` even when `unordered_code` is empty. Zero-arity `(-)` and `(/)` are legal and now reach this branch, so `begin(unordered_code) + 1` is outside the valid iterator range passed to `std::sort` (undefined behavior). A release probe happened to print `(-)` and `(/)`, but that does not make the iterator arithmetic valid. Please clamp the preserved prefix to `unordered_code.size()` (or sort only when `size() > starting_index`) and add sorted-unparse regressions for both zero-arity forms, preferably under debug iterators or ASan/UBSan.

src/Amalgam/evaluablenode/EvaluableNodeTreeAlgebra.cpp
```diff
@@ -0,0 +1,999 @@
...
+			currently_accumulating_nonimmediate = false;
+			nonimmediate_accum_exponent = 0.0;
+		}
+
+		//add on accumulated_immediate at end if nonzero
+		if(product_immediate != 1.0)
+		{
+			if(ocn.size() > 0)
+			{
+				ocn.push_back(rc.enm->AllocNode(division ? 1.0 / product_immediate : product_immediate));
```
howso-mario: Blocking for the stated canonicalization/mergeability goal: division tails are accumulated as reciprocals and inverted again when emitted, so a simple one-constant rewrite perturbs the literal. Current HEAD runtime results:

```text
(simplify (lambda (/ a 49))) -> (/ a 49.00000000000001)
(call (simplify (lambda (/ a 7 7))) {a 49}) -> 0.9999999999999999
```

This is not a complaint about preserving IEEE operation order; the same divisor is round-tripped through `1/x` and changed, which harms stable code equality and changes a direct call result. Accumulate denominator-position immediates multiplicatively and emit that product directly; only invert in the fully collapsed literal-result branch. Add regressions for `(/ a 49)`, `(/ a 7 7)`, the `a=49` call, and retain zero-divisor coverage.

howso-mario (changes requested):
Requesting changes on current head `1a840713` after a fresh full-diff review against current `main`. I found three focused issues in the new simplify/sort behavior; the inline comments include exact reproducers and patch/test shapes.

Verification: the CI-faithful `amd64-release-linux-228` configure/build/test completed all 472 build steps and passed 13/13 CTest smoke targets. Remote PR checks are also green across all product builds/smoke tests/docs regeneration. Focused runtime probes on that build nevertheless returned `"adXc"` for an ordered concat whose original result is `"aXcd"`, rewrote `(/ a 49)` to `(/ a 49.00000000000001)`, and returned `0.9999999999999999` for the corresponding `a=49` call. The zero-arity sorted-unparse case happens to print successfully in this release build, but constructs an invalid random-access iterator range and is undefined behavior.

I inventoried all six historical threads before reviewing current HEAD. The null-result crash, boolean absorbing value, distinguished subtraction/division operand/arity, and generic `Merger<T>` issues are fixed and are not restated. The prior division thread is only partially superseded because the reciprocal round-trip below remains. Two independent review lanes (Claude Code primary, Codex second-opinion) agreed on the ordered-concat and invalid-sort-range blockers; both also identified the reciprocal canonicalization defect. No other blocking findings remain from this pass.

src/Amalgam/evaluablenode/EvaluableNodeTreeAlgebra.cpp
howso-mario: Fixed in d5ed78b by splicing each nested associative node at its original operand index and rechecking inserted children. Added one- and two-nested-concat regressions (`aXcd` and `aXmYd`). GCC 10 container build, all 13 CTest targets, full `--validate-amalgam`, and `git diff --check` pass.

src/Amalgam/Parser.cpp
howso-mario: Fixed in d5ed78b by clamping the ordered-prefix index to the generated child-code vector size before iterator arithmetic. Added zero-, one-, and multi-argument unparse coverage; the multi-argument case also verifies only the unordered suffix is sorted. GCC 10 container build, all 13 CTest targets, full `--validate-amalgam`, and `git diff --check` pass.

src/Amalgam/evaluablenode/EvaluableNodeTreeAlgebra.cpp
howso-mario: Fixed in d5ed78b by accumulating division-tail literals as a direct denominator product and emitting that product once, avoiding reciprocal round-trips. Added exact 49, evaluation-to-1, second-pass idempotence, zero/signed-zero, and multi-divisor regressions. GCC 10 container build, all 13 CTest targets, full `--validate-amalgam`, and `git diff --check` pass.

src/Amalgam/evaluablenode/EvaluableNodeTreeAlgebra.cpp
```diff
@@ -0,0 +1,1003 @@
...
+		return (child_structure_type == OpcodeDetails::ChildNodeStructureType::UNORDERED ||
+				child_structure_type == OpcodeDetails::ChildNodeStructureType::ONE_POSITION_THEN_UNORDERED_OR_ONE_UNORDERED);
+	}
+
+	bool Rewrite(EvaluableNode *en, RuleContext &rc)
+	{
+		auto &ocn = en->GetOrderedChildNodesReference();
+		if(GetChildNodeStructureType(en->GetType()) == OpcodeDetails::ChildNodeStructureType::UNORDERED)
+			std::sort(begin(ocn), end(ocn), EvaluableNode::IsStrictlyLessThan);
+		else if(ocn.size() > 2)
```
howso-mario, line 842: Blocking: sorting the divisor tail is unsafe when a zero divisor is present because runtime division stops at the first zero. Current head rewrites `(/ a 0 -1)` to `(/ a -1 0)`; in all five tested CI-image binaries, the original with `a=2` returned `.infinity` while the simplified form returned `-.infinity`. It can also move an effectful operand before a zero that previously prevented its evaluation. Please preserve divisor order through/after the first zero (or otherwise avoid this rewrite) and add regressions for `(/ a 0 -1)` and `(/ a 0 -2)`.

src/Amalgam/evaluablenode/EvaluableNodeTreeAlgebra.cpp
```diff
@@ -0,0 +1,1003 @@
...
+			return false;
+
+		EvaluableNode *child = ocn[0];
+		if(EvaluableNode::IsNull(child))
+		{
+			rc.NullifyNode(en);
+			return true;
+		}
+
+		if(child->GetType() == ENT_LOG)
```
howso-mario: Blocking: `exp(log(x)) -> x` is valid only for positive-real inputs, and the analogous `log(exp(x)) -> x` rule below is not valid after `exp` overflows. CI-image probes across five binaries reproduced `exp(log(-1))` as `.null` versus simplified `-1`, and `log(exp(1000))` as `.infinity` versus simplified `1000`. This conflicts with `simplify`'s same-behavior/same-result contract. Please remove these unconditional identities or require a domain/range proof, with focused negative/zero/infinity/overflow regressions.

src/Amalgam/evaluablenode/EvaluableNodeTreeAlgebra.cpp
```diff
@@ -0,0 +1,1003 @@
...
+						else
+							nonimmediate_accum_multiplicand -= 1.0;
+					}
+					else
+					{
+						currently_accumulating_nonimmediate = true;
+						if(!subtraction)
+							nonimmediate_accum_multiplicand = 2.0;
+						else
+						{
```
howso-mario: Blocking: cancelling equal symbolic operands assumes a finite numeric binding. Current head folds `(- a a)` to `0` and the analogous division rule folds `(/ a a)` to `1`; focused probes across five binaries showed the originals return `.null` for `.null`/`.infinity` bindings while simplified forms return `0`/`1`. Please restrict these cancellations to values proven safe at simplify time (including division's nonzero/finite requirement), or retain the executable expression, and add null/infinity/NaN regressions.

src/Amalgam/evaluablenode/EvaluableNodeTreeAlgebra.cpp
```diff
@@ -0,0 +1,1003 @@
...
+				i -= 2;
+			}
+		}
+
+		//check for low number of parameters
+		if(!node_replaced)
+		{
+			if(ocn.size() == 0)
+			{
+				rc.NullifyNode(en);
```
howso-mario, line 173: Blocking UB: `ocn` refers to the node's ordered-child vector, but `NullifyNode(en)` changes the node type and destroys that vector; the next `ocn.size()` reads a destructed object. Please return immediately after nullifying (as the multiplication/division fold already does). The same lifetime bug exists in `FoldAdditionAndSubtraction` after `SetTypeViaNumberValue` and before the later `ocn.size()` check. A debug-iterator or sanitizer regression around zero-arity `if` and all-immediate folding would make this durable.

howso-mario (changes requested):
CHANGES_REQUESTED at `6474cece`. CI-faithful verification passed: `ghcr.io/howsoai/amalgam-build-container-linux-228:2.0.9`, preset `amd64-release-linux-228`, configure + full 472-target build + CTest 13/13. Live GitHub checks are 19 successful and 5 expected release-only skips. Focused probes on ST, ST-PGC, ST-AFMI, MT, and MT-AFMI reproduced the semantic differences in the inline blockers.

src/Amalgam/evaluablenode/EvaluableNodeTreeAlgebra.cpp
howso-mario: FIXED: Fixed in e4d0ce1. Division sorting now stops before the first literal zero divisor, preserves the zero and every later operand in original order, and skips multiplication/division folding whenever a literal zero divisor is present. Regressions cover `(/ a 0 -1)`, `(/ a 0 -2)`, multiple divisors around zero, and observed infinity sign. Verified with the GCC 10 / glibc 2.28 CI image: full build and 13/13 CTest targets pass.

src/Amalgam/evaluablenode/EvaluableNodeTreeAlgebra.cpp
howso-mario: FIXED: Fixed in e4d0ce1 by removing both unconditional inverse rewrites (`exp(log(x))` and `log(exp(x))`). Exact `ln(e) -> 1` and null propagation remain. Regressions cover negative, zero, infinity, overflow, underflow, and structural forms. Verified with the GCC 10 / glibc 2.28 CI image: full build and 13/13 CTest targets pass.

src/Amalgam/evaluablenode/EvaluableNodeTreeAlgebra.cpp
howso-mario: FIXED: Fixed in e4d0ce1. Duplicate grouping for subtraction/division now excludes the distinguished head operand, so symbolic `(- a a)` and `(/ a a)` remain executable and preserve null/infinity/zero behavior; non-head grouping remains enabled. NaN is normalized to `.null` by node allocation, so stable regression coverage uses null plus infinity/zero bindings. Verified with the GCC 10 / glibc 2.28 CI image: full build and 13/13 CTest targets pass.

src/Amalgam/evaluablenode/EvaluableNodeTreeAlgebra.cpp
howso-mario: FIXED: Fixed in e4d0ce1. `DeadCodeElimination` now returns immediately after `NullifyNode`, and `FoldAdditionAndSubtraction` assigns the accumulated value and returns without reading the destroyed ordered-child vector after a type change. Regression examples cover zero-arity `if`, all-immediate folding, and simplify-node mutation paths; an ASan+UBSan validation pass also completed all 959 examples with no sanitizer errors. The GCC 10 / glibc 2.28 CI-image build and all 13 CTest targets pass.

howso-mario (approved):
Re-reviewed current head 39a12b1 against the previously verified bd5e2ae head. The four findings from review 4912005498 remain resolved: zero-divisor ordering is preserved, unsafe exp/log inverses remain removed, symbolic subtraction/division heads are not cancelled, and node-type mutation paths return before reusing invalidated child vectors. The post-review delta is limited to comment cleanup, a clean merge of main’s C++20/build update, and consistent Visual C++ stdcpp20 settings. All current required CI checks, including Windows, GCC 14 Linux/glibc 2.28, ARM, macOS, WASM, docs, smoke tests, and final-check, are green; release-only jobs are skipped as expected.

howsoRes (approved):

PR #636, opened 2026-08-13
15994: Improves code to take advantage of C++20 capabilities, improves performance

howsohazard:
* Improves code to take advantage of C++20 capabilities, improves performance (especially via [[likely]] / [[unlikely]] tags)
* Updates simdjson library

### DIFF
--- a/src/3rd_party/simdjson/simdjson.h
+++ b/src/3rd_party/simdjson/simdjson.h
@@ -1,4 +1,4 @@
-/* auto-generated on 2026-05-06 17:28:39 -0400. version 4.6.4 Do not edit! */
+/* auto-generated on 2026-07-30 16:20:13 -0400. version 4.6.6 Do not edit! */
 /* including simdjson.h:  */
 /* begin file simdjson.h */
 #ifndef SIMDJSON_H
@@ -2538,7 +2538,7 @@ namespace std {
 #define SIMDJSON_SIMDJSON_VERSION_H
 
 /** The version of simdjson being used (major.minor.revision) */
-#define SIMDJSON_VERSION "4.6.4"
+#define SIMDJSON_VERSION "4.6.6"
 
 namespace simdjson {
 enum {
@@ -2553,7 +2553,7 @@ enum {
   /**
    * The revision (major.minor.REVISION) of simdjson being used.
    */
-  SIMDJSON_VERSION_REVISION = 4
+  SIMDJSON_VERSION_REVISION = 6
 };
 } // namespace simdjson
 
@@ -8067,16 +8067,28 @@ inline std::pair<std::string_view, std::string_view> get_next_key_and_json_path(
     }
 
     key = json_path.substr(key_start, i - key_start);
-  } else if ((i+1 < json_path.size()) && json_path[i] == '[' && (json_path[i+1] == '\'' || json_path[i+1] == '"')) {
+  } else if ((i + 1 < json_path.size()) && json_path[i] == '[' &&
+             (json_path[i + 1] == '\'' || json_path[i + 1] == '"')) {
+    // Bracket-quoted key: ['key'] or ["key"].
+    // Require a matching closing quote and a following ']'. If either is
+    // missing, return an empty key and the original path so callers can treat
+    // this as a parse failure (e.g. INVALID_JSON_POINTER) without advancing.
+    // Without this check, i += 2 can make i > size() and substr throws
+    // std::out_of_range, which aborts noexcept callers such as
+    // at_path_with_wildcard / for_each_at_path_with_wildcard.
+    const char quote = json_path[i + 1];
     i += 2;
-    size_t key_start = i;
-    while (i < json_path.length() && json_path[i] != '\'' && json_path[i] != '"') {
+    const size_t key_start = i;
+    while (i < json_path.length() && json_path[i] != quote) {
       ++i;
     }
-
+    if (i >= json_path.length() ||               // missing closing quote
+        i + 1 >= json_path.length() ||           // missing ]
+        json_path[i + 1] != ']') {
+      return {key, json_path};
+    }
     key = json_path.substr(key_start, i - key_start);
-
-    i += 2;
+    i += 2; // past quote and ]
   } else if ((i+2 < json_path.size()) && json_path[i] == '[' && json_path[i+1] == '*' && json_path[i+2] == ']') { // i.e [*].additional_keys or [*]["additional_keys"]
     key = "*";
     i += 3;
@@ -68653,6 +68665,12 @@ inline error_code array::for_each_at_path_with_wildcard(std::string_view json_pa
   auto result_pair = get_next_key_and_json_path(json_path);
   std::string_view key = result_pair.first;
   std::string_view remaining_path = result_pair.second;
+  // Empty key means the path segment could not be parsed (including malformed
+  // bracket-quoted keys). Without this check, an empty key is treated as index
+  // 0 and remaining_path may be unchanged, causing infinite recursion.
+  if (key.empty()) {
+    return INVALID_JSON_POINTER;
+  }
   // Wildcard case
   if (key=="*"){
     for(auto element: *this) {
@@ -72182,6 +72200,12 @@ inline error_code object::for_each_at_path_with_wildcard(std::string_view json_p
   auto result_pair = get_next_key_and_json_path(json_path);
   std::string_view key = result_pair.first;
   std::string_view remaining_path = result_pair.second;
+  // Empty key means the path segment could not be parsed (including malformed
+  // bracket-quoted keys). Match DOM at_path_with_wildcard and avoid treating
+  // "no progress" as a recursive lookup of "".
+  if (key.empty()) {
+    return INVALID_JSON_POINTER;
+  }
   // Handle when its the case for wildcard
   if (key == "*") {
     // Loop through each field in the object
@@ -72547,7 +72571,7 @@ simdjson_inline parser::parser(size_t max_capacity) noexcept
 
 simdjson_warn_unused simdjson_inline error_code parser::allocate(size_t new_capacity, size_t new_max_depth) noexcept {
   if (new_capacity > max_capacity()) { return CAPACITY; }
-  if (string_buf && new_capacity == capacity() && new_max_depth == max_depth()) { return SUCCESS; }
+  if (string_buf && new_capacity <= capacity() && new_max_depth <= max_depth()) { return SUCCESS; }
 
   // string_capacity copied from document::allocate
   _capacity = 0;
@@ -82023,6 +82047,12 @@ inline error_code array::for_each_at_path_with_wildcard(std::string_view json_pa
   auto result_pair = get_next_key_and_json_path(json_path);
   std::string_view key = result_pair.first;
   std::string_view remaining_path = result_pair.second;
+  // Empty key means the path segment could not be parsed (including malformed
+  // bracket-quoted keys). Without this check, an empty key is treated as index
+  // 0 and remaining_path may be unchanged, causing infinite recursion.
+  if (key.empty()) {
+    return INVALID_JSON_POINTER;
+  }
   // Wildcard case
   if (key=="*"){
     for(auto element: *this) {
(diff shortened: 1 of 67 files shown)


src/Amalgam/AmalgamAPI.cpp
```diff
@@ -84,12 +67,12 @@ extern "C"
 			char *entity_path_alloc = new char[sizeof(char *) * entity_path_len];
 			entity_path = reinterpret_cast<char **>(entity_path_alloc);
 			for(size_t i = 0; i < entity_path_len; i++)
-				entity_path[i] = StringToCharPtr(status.entity_path[i]);
+				entity_path[i] = strdup(status.entity_path[i].c_str());
```
howso-mario: [blocking] `strdup` allocates with `malloc`, but the documented public deallocator `DeleteString` still uses `delete[]` (`AmalgamAPI.cpp:305-308`). That allocator mismatch is undefined behavior and can be caught as memory corruption by ASan/MSVC debug runtimes. The same replacement appears for every returned narrow string, while other API buffers still use `new char[]`, so changing only `DeleteString` to `free` would break those. Please keep one allocation family end-to-end: either restore a `new char[]` copy helper for these call sites, or convert all buffers handled by `DeleteString` to malloc/free consistently. This also needs the portable `_strdup` spelling on MSVC if strdup remains.

src/Amalgam/PlatformSpecific.h
```diff
@@ -156,40 +88,73 @@ inline std::pair<std::string, bool> Platform_OpenFileAsString(const std::string
...
-	std::string stringified_s(s);
-	const char *start_pointer = stringified_s.data();
+	// FALLBACK FOR APPLECLANG, WASM, AND OLDER PLATFORMS
+
+	const char *start_pointer = s.data();
+
+	// Fast safety check: if s is a type like std::string, s[s.size()] is guaranteed 
+	// to be '\0' in C++11 and later. If it's a slice or view, we must copy it.
+	std::string fallback_copy;
+	if(start_pointer[s.size()] != '\0')
```
howso-mario: [blocking] This probes `data()[size()]` to determine whether an arbitrary `StringType` is terminated, but that read is out of bounds for a `std::string_view` whose view reaches the end of its backing allocation/object. Real callers pass intern-pool views and sliced views (for example `Parser.cpp:207-209`), so the AppleClang/WASM fallback can perform an invalid read. Please do not inspect one-past-the-view; copy view-like inputs into a `std::string` (or use an overload/type constraint so only a known `std::string` skips the copy).

src/Amalgam/PlatformSpecific.h
```diff
@@ -45,82 +45,14 @@
...
-#else
-	return false;
-#endif
-//Windows is almost exclusively little endian, even on ARM
-#elif defined(_MSC_VER)
-	return true;
+//ensure strdup is portable
+#if defined(_MSC_VER)
+#define PORTABLE_STRDUP _strdup
+#pragma warning(push)
```
howso-mario: This `#pragma warning(push)` has no matching `#pragma warning(pop)`, so including this header leaves C4996 disabled for the rest of every MSVC translation unit. Please remove the pragma if `_strdup` avoids the warning, or balance it immediately around the intended code.

src/Amalgam/interpreter/OpcodesBasicMath.cpp
```diff
@@ -671,8 +671,8 @@ EvaluableNodeReference Interpreter::InterpretNode_ENT_SET_DIGITS(EvaluableNode *
 {
 	auto &ocn = en->GetOrderedChildNodesReference();
 	size_t num_params = ocn.size();
-	if(num_params == 0)
-		return AllocReturn(std::numeric_limits<double>::quiet_NaN(), immediate_result);
+	if(num_params == 0) [[unlikely]]
+		return EvaluableNodeReference::Null();
```
howso-mario, line 675: This changes `(set_digits)` with no arguments from returning NaN to returning null, which is a user-visible language semantic change rather than mechanical C++20 cleanup. The docs currently say this opcode returns a number, and this PR has no corresponding docs/test update. If intentional, please document and test it; otherwise retain the previous NaN behavior.

howso-mario (changes requested):
Review of exact head `3e5a132b13a11cf4e0850612c4a83f3a91d11fe3`. The mechanical C++20 cleanups generally look sound, and all current CI/build/smoke-test checks are green across Linux, glibc 2.28, ARM, macOS, Windows, and WASM. I found two high-severity correctness issues that need fixes before merge: a `strdup`/`delete[]` allocator mismatch in the public C API, and an out-of-bounds read in the AppleClang/WASM number-parsing fallback. I also left comments for the unbalanced MSVC warning pragma and an undocumented `set_digits` behavior change.

I additionally checked the ranges key/value conversions and lifetimes, bit-operation replacements, `std::endian`, `std::atomic_ref`, ThreadPool rewrite, simdjson 4.6.6 amalgamation update, and brace-return conversions. No other blocking issue found there. `git diff --check` reports one trailing-space error in `PlatformSpecific.h:121`.

src/Amalgam/interpreter/OpcodesBasicMath.cpp
howsohazard: NaNs are nulls in Amalgam and have no independent representation.  So therefore the behavior is identical but it is more efficient rather than AllocReturn checking if it is a NaN and then returning a null.

src/Amalgam/AmalgamAPI.cpp
```diff
@@ -321,7 +304,7 @@ extern "C"
 
 	void DeleteString(char *p)
 	{
-		delete[] p;
+		free(p);
```
howso-mario: [blocking] This fixes the `strdup` half of the prior allocator mismatch but breaks the other half exactly as the earlier review warned: `DeleteString` is also the documented deallocator for buffers that this file still allocates with `new[]`—the `LoadEntityStatus::entity_path` outer array (`AmalgamAPI.cpp:67`, contract in `Amalgam.h:25-27`), `StoreEntityToMemory` (`AmalgamAPI.cpp:170`, whose comment directs callers to `DeleteString`), and both levels returned by `GetEntities` (`AmalgamAPI.cpp:290,294`). Passing any of those to `free` is undefined behavior. Please preserve one allocation family end-to-end. The minimal safe patch is to restore `delete[]` here and replace every `strdup(...)` call in this file with a `new[]` helper:

```cpp
char *StringToCharPtr(const std::string &value)
{
	char *out = new char[value.size() + 1];
	std::memcpy(out, value.data(), value.size());
	out[value.size()] = '\0';
	return out;
}
```

Then use `StringToCharPtr(status.entity_path[i])`, `StringToCharPtr(status.message)`, etc., and keep:

```cpp
void DeleteString(char *p)
{
	delete[] p;
}
```

This also avoids relying on non-standard `strdup` spellings. The alternative is to convert *every* buffer documented for `DeleteString` (including the outer arrays) to `malloc` consistently, but that is a broader and easier-to-miss ABI change. Please add cleanup coverage for `StoreEntityToMemory`, `LoadEntityStatus::entity_path`, and `GetEntities`; the existing smoke tests exercise only some `strdup`-backed returns and therefore stay green despite this UB.

src/Amalgam/PlatformSpecific.h
```diff
@@ -45,82 +45,13 @@
...
-#else
-	return false;
-#endif
-//Windows is almost exclusively little endian, even on ARM
-#elif defined(_MSC_VER)
-	return true;
+//ensure strdup is portable
+#if defined(_MSC_VER)
+#define PORTABLE_STRDUP _strdup
+#pragma warning(disable : 4996)
```
howso-mario: [blocking portability/diagnostics] Removing the unmatched `push` did not scope the warning suppression; it made C4996 permanently disabled from this header onward in every MSVC translation unit. The macro is currently unused because `AmalgamAPI.cpp` calls `strdup` directly, and `_strdup` itself does not require suppressing C4996. With the `new[]` helper proposed in the allocator comment, remove this entire `PORTABLE_STRDUP` block. If the macro is retained, the exact replacement is:

```cpp
#if defined(_MSC_VER)
#define PORTABLE_STRDUP _strdup
#else
#define PORTABLE_STRDUP strdup
#endif
```

That fixes the diagnostic leak; every retained call would also need to use `PORTABLE_STRDUP` rather than `strdup` directly.

howso-mario (changes requested):
Re-review of exact head `a3d29a14174853f5f113608ffccd813b4aee816c`. The prior out-of-bounds string-view probe is fixed by the unconditional owned `std::string` copy, and the author's `set_digits` explanation is correct: `AllocReturn(NaN)` already produces null in both immediate and allocated representations. The cumulative 66-file diff and surrounding lifetime/bounds/portability code were re-read, and the current CI matrix is green across Linux, glibc 2.28, ARM, macOS, Windows, WASM, smoke tests, docs, and final-check.

The allocator blocker is not fixed end-to-end: changing only `DeleteString` to `free` makes the still-`new[]` public API buffers undefined behavior when callers follow the documented deallocation contract. The MSVC header also still disables C4996 for every including translation unit. See the two inline comments for exact replacement code. `git diff --check` is clean. I did not modify the contributor branch.

src/3rd_party/simdjson/simdjson.h
```diff
@@ -68653,6 +68665,12 @@ inline error_code array::for_each_at_path_with_wildcard(std::string_view json_pa
   auto result_pair = get_next_key_and_json_path(json_path);
   std::string_view key = result_pair.first;
   std::string_view remaining_path = result_pair.second;
+  // Empty key means the path segment could not be parsed (including malformed
+  // bracket-quoted keys). Without this check, an empty key is treated as index
+  // 0 and remaining_path may be unchanged, causing infinite recursion.
+  if (key.empty()) {
```
howso-mario, line 68671: [blocking correctness] This guard conflates an empty parsed key with parser failure. A valid JSONPath such as `['']` against `{"": 1}` produces an empty `key` *and* consumes the whole path; before this change the object loop could match that empty member name, but all 18 architecture copies now return `INVALID_JSON_POINTER`. Detect the actual malformed-path signal instead: the helper returns the original, unadvanced view on failure. Replace every newly added `key.empty()` guard with:

```cpp
if (remaining_path.data() == json_path.data() &&
    remaining_path.size() == json_path.size()) {
  return INVALID_JSON_POINTER;
}
```

This preserves valid empty keys while still stopping recursion for unterminated bracket-quoted paths. Apply the equivalent replacement to each generated architecture copy and add focused coverage for `{"":1}` + `['']` (success), `['x` and `['x'` (invalid without recursion), and malformed/empty array indices. The bounds checks in `get_next_key_and_json_path` itself are correct; only this failure predicate needs changing.

howso-mario (changes requested):
Final re-review of exact head `889087f56aa5b29bfa7bfc344897bdb3707b4ad0`. The two blockers from the previous round are fixed: all documented narrow `DeleteString` buffers are again allocated with `new[]` and released with `delete[]`, and C4996 suppression is now balanced locally around `wcsncpy`. The prior string-view parsing fallback and no-argument `set_digits` concerns also remain resolved. However, the new simdjson malformed-path guard introduces a correctness regression for valid empty object keys, so I cannot approve this revision. See the inline finding for an exact replacement.

Verification: re-read the cumulative 66-file diff and newest incremental diff; `git diff --check` passes; CI is fully green for Linux amd64/glibc-2.28/ARM, macOS ARM64, Windows amd64, WASM debug/release, smoke tests, docs, and final-check; CI-faithful local glibc-2.28 container build completed and CTest passed 13/13 (shared-library smoke suites each 67/67). No contributor files were modified.

src/3rd_party/simdjson/simdjson.h
howsohazard: This is a 3rd party library, and this code is generated by their automated process.  Won't fix, and will wait for newer version of their library, or we can revert to the older version and wait for the next update.

howso-mario (approved):
Approving exact head `889087f56aa5b29bfa7bfc344897bdb3707b4ad0`. The previously reported empty-object-key JSONPath behavior is in generated third-party simdjson code across the architecture copies and is accepted for this dependency update as a known upstream limitation; it should not be locally hand-patched. The prior first-party allocator, warning-scope, string-view, and `set_digits` blockers remain fixed, and the current required CI matrix is green.

